PR #12, opened 2023-07-06
Add Japanese README

eltociear:
I created Japanese translated README.


geekan:
@eltociear thx for your contribution. but i doubt it's difficult to maintain a Thank you very much for your contribution, but the current README is still changing frequently, and I am worried that this will cause difficulties in subsequent maintenance. Not sure if you want to participate in follow-up maintenance?

eltociear:
@geekan I will update the Japanese version when README.md is updated.

PR #16, opened 2023-07-07
Adding a Dockerfile for docker implementation

sablin39:
I wrote a Dockerfile that can use the same config as the host machine and stores the results in the host as well.


jmanhype:
Does this work?

jmanhype:
Working?

geekan:
Requires optimization. Minimize the size of the image as much as possible, and try to add the conditional judgment for selecting the source.

sablin39:
I add some cache removal commands to reduce the size of the image. Moreover, I've push it to dockerhub so you can simply pull and try the image, have fun :)

PR #29, opened 2023-07-10
Optimize Dockerfile and README

voidking:


voidking:
1. The size of the image is reduced from 9.9G to 7.2G
2. Concrete commands using Docker in README

PR #37, opened 2023-07-12
2023 07 10 chengmaoyu

qa6300525:
2.2 Establish a separate output queue, differentiated from the message queue

1. Validate LLM output format and retry if an error occurs.
2. Add a new standard output format.
3. Ensure consistency between the standard format definition and what's in the prompt.
4. Implement a one-shot mechanism to assist LLM in producing more accurate outputs.

For details, see the document    https://docs.qq.com/doc/DSGVMdE9lRVJEb0RH

### DIFF
--- a/metagpt/actions/__init__.py
+++ b/metagpt/actions/__init__.py
@@ -8,6 +8,7 @@
 from enum import Enum
 
 from metagpt.actions.action import Action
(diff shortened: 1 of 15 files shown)


drnic:
Question: why did you name the method `_aask_v1` -- what might a better name be to describe what it does, and how it is different from `_aask`?

PR #61, opened 2023-07-21
add long-term memory serialization&deserialization

better629:
1. add long-term memory serialization&deserialization
2. add unittest of serialization&deserialization
3. re-startup with same idea will return empty due to current role `_observe` strategy. 
3.1 TODO：It should run when re-startup with same idea but reuse the role's existed memory


geekan:
This implementation is closed to "cache" feature, not longterm memory.
The cache feature should use pydantic, and avoid customize Serialization / Deserialization functions.

Longterm memory should use embedding-search function to get similar memory.

qa6300525:
The basic functionality doesn't have issues, but the overall system is not complete yet.

The current memory system saves project conversation records without differentiating ideas. This results in any task pulling from the historical memory, leading to a muddled context.
The purpose of the long-term memory may need to be reevaluated. My understanding is that it should be used during incremental tasks or for task recovery (suggest focusing on task recovery first).
Therefore, I recommend resolving the above two points before merging.

基础功能没有问题，但是整体还不完整。

1. 当前的记忆是保存的项目对话记录，并且没有区分idea，因此任何任务都会读取历史记忆，会造成context混杂。
2. 长记忆可能需要再评估下是在做什么？我的理解是在增量任务的时候使用 或者 任务恢复的时候使用（建议先做任务恢复）？

因此建议需要先解决以上两点再合入。

PR #62, opened 2023-07-21
[bugfix] fix serper integration bug to support batch queries

LeonZh0u:


geekan:
If met "a|b" query, the implementation would be wrong.

voidking:
This pr is behind the main branch a lot. Please rebase it first.

PR #70, opened 2023-07-24
add memory_storage using ann to avoid similar idea repetitive execution

better629:
1. Use FaissStore as MemoryStorage to store the watched action's related messages of a role.
2. Startup with similar idea will return with `no news` in log directly due to the existed memory.

### DIFF
--- a/metagpt/roles/role.py
+++ b/metagpt/roles/role.py
@@ -12,10 +12,11 @@ from typing import Iterable, Type
 from pydantic import BaseModel, Field
 
 # from metagpt.environment import Environment
+from metagpt.config import CONFIG
 from metagpt.actions import Action, ActionOutput
 from metagpt.llm import LLM
 from metagpt.logs import logger
-from metagpt.memory import Memory
+from metagpt.memory import Memory, LongTermMemory
 from metagpt.schema import Message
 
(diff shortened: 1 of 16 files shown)


metagpt/roles/role.py
```diff
@@ -174,13 +184,7 @@ async def _observe(self) -> int:
...
         observed = self._rc.env.memory.get_by_actions(self._rc.watch)
         
-        already_observed = self._rc.memory.get()
-        
-        news: list[Message] = []
-        for i in observed:
-            if i in already_observed:
-                continue
-            news.append(i)
+        news = self._rc.memory.remember(observed)  # remember recent exact or similar memories
```
geekan, line 187: remember一般会同时存在长期和短期记忆，短期记忆比如最近10个

better629: > remember一般会同时存在长期和短期记忆，短期记忆比如最近10个

@geekan Do you means that remember from short-term mem first and then integrate with the long-term result ?

better629: @geekan integrate stm & ltm with latest commit

shenchucheng (approved):
LGTM

geekan:
LGTM

PR #96, opened 2023-07-30
Move puppeteer-config.json to config directory.

voidking:


config/config.yaml
```diff
@@ -2,10 +2,10 @@
 # The configuration of key.yaml has a higher priority and will not enter git
 
 #### if OpenAI
-
-#OPENAI_API_KEY: "YOUR_API_KEY"
-#OPENAI_API_BASE: "YOUR_API_BASE"
+## You can get OPENAI_API_BASE url from https://github.com/beidongjiedeguang/openai-forward
+OPENAI_API_BASE: "https://api.openai-forward.com/v1"
```
shenchucheng, line 6: Is it better to add `#` here?

config/config.yaml
```diff
@@ -2,10 +2,10 @@
...
 #OPENAI_PROXY: "http://127.0.0.1:8118"
```
shenchucheng, line 7: `OPENAI_PROXY` is a variable used to configure a proxy for OpenAI API requests when there is a need to access `OPENAI_API_BASE` through a proxy server. It may be helpful to reduce user confusion by clarifying the purpose of OPENAI_PROXY.

config/config.yaml
voidking: Give user a default available value, so that user only needs to modify OPENAI_API_KEY and OPENAI_API_MODEL.

shenchucheng:
LGTM

PR #112, opened 2023-08-04
optimize logging for network issues.

wangjie5540:
reason: Some guys meet the network issue like "443 error, Connect error"

@geekan PTAL


metagpt/config.py
```diff
@@ -46,14 +46,18 @@ def __init__(self, yaml_file=default_yaml_file):
...
         if not self.openai_api_base or "YOUR_API_BASE" == self.openai_api_base:
             openai_proxy = self._get("OPENAI_PROXY") or self.global_proxy
             if openai_proxy:
                 openai.proxy = openai_proxy
             else:
-                logger.info("Set OPENAI_API_BASE in case of network issues")
+                logger.warning(
+                    """
+Set OPENAI_API_BASE in case of network issues. 
+If you meet network issues, go to https://deepwisdom.feishu.cn/wiki/MsGnwQBjiif9c3koSJNcYaoSnu4#part-XdatdVlhEojeAfxaaEZcMV3ZniQ, see FAQ 5.8 "PRD卡住/无法访问/连接中断"
```
geekan, line 58: 目前默认填写了openai官方api，这个逻辑走不进来。

geekan: 除此之外，需要考虑“正常用户不显示warning”，这里严谨一些，或许在流程中直接curl目的（带超时）会好点

wangjie5540: > 除此之外，需要考虑“正常用户不显示warning”，这里严谨一些，或许在流程中直接curl目的（带超时）会好点

这一点的确是，正常能直接使用openai官网api的情况下显示warning的确不太好。

wangjie5540: 在命令行，理论上curl能通，那么MetaGPT就可以正常工作，合理的给出这样的流程性提示，感觉可以减少使用者的一些困惑。

wangjie5540:
@geekan 
在切断魔法上网以后的自测效果如下：
image

metagpt/provider/openai_api.py
```diff
@@ -150,6 +149,7 @@ def __init_openai(self, config):
     async def _achat_completion_stream(self, messages: list[dict]) -> str:
         response = await openai.ChatCompletion.acreate(
             **self._cons_kwargs(messages),
+            timeout=3,
```
geekan, line 152: put this into _cons_kwargs

metagpt/provider/openai_api.py
```diff
@@ -127,10 +116,20 @@ def get_costs(self) -> Costs:
         return Costs(self.total_prompt_tokens, self.total_completion_tokens, self.total_cost, self.total_budget)
 
 
+def log_and_reraise(retry_state):
+    logger.error("Retry attempts exhausted. Last exception: %s", retry_state.outcome.exception())
+    logger.warning("""
+Recommend going to https://deepwisdom.feishu.cn/wiki/MsGnwQBjiif9c3koSJNcYaoSnu4#part-XdatdVlhEojeAfxaaEZcMV3ZniQ
+See FAQ 5.8 PRD卡住/无法访问/连接中断
```
geekan, line 123: use english.

metagpt/provider/openai_api.py
```diff
@@ -211,7 +211,14 @@ async def acompletion(self, messages: list[dict]) -> dict:
         #     messages = self.messages_to_dict(messages)
         return await self._achat_completion(messages)
 
-    @retry(max_retries=6)
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(1),
+        after=after_log(logger, logging.WARNING),
+        retry=retry_if_exception_type(APIConnectionError),
+        reraise=True,
```
flyi, line 219: Why to raraise the exception, since you've already handled this exception? For regular users, seeing a long string of error output might lead them to believe there's a problem with the program, overlooking the possibility of the network issue.

metagpt/provider/openai_api.py
wangjie5540: ok, I will only retain the index~

metagpt/provider/openai_api.py
wangjie5540: > put this into _cons_kwargs

Thanks for reviewing, but is there necessary to put timeout into _cons_kwargs. In my opinion, _cons_kwargs will store params like temperature or model type and so on.

wangjie5540: or is this will be done?
image

metagpt/provider/openai_api.py
```diff
@@ -127,10 +116,20 @@ def get_costs(self) -> Costs:
         return Costs(self.total_prompt_tokens, self.total_completion_tokens, self.total_cost, self.total_budget)
 
 
+def log_and_reraise(retry_state):
+    logger.error("Retry attempts exhausted. Last exception: %s", retry_state.outcome.exception())
```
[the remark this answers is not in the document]
wangjie5540: You are right, I will fix it~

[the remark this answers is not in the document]
wangjie5540: > Error string formatting syntax, it's fine for logging.logger, but not loguru.logger. You can find this error in your screenshot above.

You are right, I will fix it~

metagpt/provider/openai_api.py
wangjie5540: > Why to raraise the exception, since you've already handled this exception? For regular users, seeing a long string of error output might lead them to believe there's a problem with the program, overlooking the possibility of the network issue.

Here I have two choice:
1. Give log and raise the exception, this will give chance to outer program to handle it.
image


2. Execute sys.exit when retry error
image

I prefer the first choice, but at the same time, I also want to see your thoughts.

flyi: I think set `reraise=False` might be better,  if you've already printed clearly what exception is causing this error through the logger. Traceback informations are helpful for developers but might not necessary for users.

wangjie5540: > I think set `reraise=False` might be better, if you've already printed clearly what exception is causing this error through the logger. Traceback informations are helpful for developers but might not necessary for users.

OK. I alreay remove the `reraise=True`, this will set reraise=False by default.
PTAL~

metagpt/provider/openai_api.py
```diff
@@ -211,7 +211,13 @@ async def acompletion(self, messages: list[dict]) -> dict:
         #     messages = self.messages_to_dict(messages)
         return await self._achat_completion(messages)
 
-    @retry(max_retries=6)
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(1),
+        after=after_log(logger, logging.WARNING),
```
better629, line 217: Maybe we only want to use loguru, can you replace logging.WARNING with `from loguru._defaults import LOGURU_WARNING_NO` (it's protected but can import)

metagpt/provider/openai_api.py
```diff
@@ -127,10 +116,20 @@ def get_costs(self) -> Costs:
         return Costs(self.total_prompt_tokens, self.total_completion_tokens, self.total_cost, self.total_budget)
 
 
+def log_and_reraise(retry_state):
+    logger.error(f"Retry attempts exhausted. Last exception: {retry_state.outcome.exception()}")
```
better629, line 120: If the `exception()` return all Traceback messages or just the last one. If it's better to use `logger.exception` to catch the `exceptions()`?

metagpt/provider/openai_api.py
wangjie5540: > Maybe we only want to use loguru, can you replace logging.WARNING with `from loguru._defaults import LOGURU_WARNING_NO` (it's protected but can import)

You are right, we must focus with `loguru`, but`from loguru._defaults import LOGURU_WARNING_NO` will cause compile warning like this:
image

So I decide to use `logger.level('WARNING').no` instead of  `from loguru._defaults import LOGURU_WARNING_NO`. 
How do you feel about that?
image

metagpt/provider/openai_api.py
wangjie5540: > If the `exception()` return all Traceback messages or just the last one. If it's better to use `logger.exception` to catch the `exceptions()`?

I prefer giving *recommend log text*, like this:
image

By the way, I am interest to test what you said, but getting a object print when I `print(logger.exception)`

metagpt/provider/openai_api.py
better629: It should be `logger.level("WARNING").no`

metagpt/provider/openai_api.py
better629: the usage is `logger.exception(f"Retry attempts exhausted. Last exception: {retry_state.outcome.exception()}")`. You can have a test if it will print the whole Traceback like below in the log file
image

metagpt/provider/openai_api.py
wangjie5540: > or is this will be done? image

It is done like above~ PTAL

metagpt/provider/openai_api.py
wangjie5540: > It should be `logger.level("WARNING").no`

I have tested two situations with `logger.level("WARNING").no` or `logger.level("WARNING").name`
this is the situation with `logger.level("WARNING").no`:
image

Maybe using `logger.level("WARNING").name` will be more suitable for normal users.

metagpt/provider/openai_api.py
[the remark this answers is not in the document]
wangjie5540: > also suggest putting it into _cons_kwargs

Done~ PTAL

metagpt/provider/openai_api.py
wangjie5540: > the usage is `logger.exception(f"Retry attempts exhausted. Last exception: {retry_state.outcome.exception()}")`. You can have a test if it will print the whole Traceback like below in the log file image

Thanks for your advicing, and I am satisfied to learn to use it. 
But as my further test, `logger.exception` has no trace stack at that time. Let's see:
image

metagpt/provider/openai_api.py
better629: In your previous code `after_log(logger, logging.WARNING)`, logging.WARNING=30, so I suggest `logger.level("WARNING").no`. It it should be `WARNING` string, you can use `logger.level("WARNING").name`.

wangjie5540: > In your previous code `after_log(logger, logging.WARNING)`, logging.WARNING=30, so I suggest `logger.level("WARNING").no`. If it should be `WARNING` string, you can use `logger.level("WARNING").name`.

OK. I have done with it. PTAL~

better629:
LGTM

stellaHSR:
LGTM

PR #127, opened 2023-08-06
Update README.md

ak8893893:
(If you don't have npm in your computer, please go to the Node.js offical website to install Node.js https://nodejs.org/ and then you will have npm tool in your computer.)


martcpp:
This is good merge

PR #138, opened 2023-08-07
remove config: update_costs

alitrack:
`CALC_USAGE` default is `true`, 
If the default `calc_usage` and `update_cost` operations fail, the subsequent execution of these two functions will be omitted.


metagpt/provider/openai_api.py
```diff
@@ -255,10 +260,15 @@ async def acompletion_batch_text(self, batch: list[list[dict]]) -> list[str]:
...
-            self._cost_manager.update_cost(prompt_tokens, completion_tokens, self.model)
+        if CONFIG.calc_usage:
+            try:
+                prompt_tokens = int(usage['prompt_tokens'])
+                completion_tokens = int(usage['completion_tokens'])
+                self._cost_manager.update_cost(prompt_tokens, completion_tokens, self.model)
+            except Exception as e:
+                logger.error("updating costs failed!", e)
+                CONFIG.calc_usage = False
+                pass
```
stellaHSR, line 271: rm this 'pass'

metagpt/provider/openai_api.py
```diff
@@ -224,13 +224,17 @@ async def acompletion_text(self, messages: list[dict], stream=False) -> str:
...
+            usage = {}
+            try:
+                prompt_tokens = count_message_tokens(messages, self.model)
+                completion_tokens = count_string_tokens(rsp, self.model)
+                usage['prompt_tokens'] = prompt_tokens
+                usage['completion_tokens'] = completion_tokens
+                return usage
+            except Exception as e:
+                logger.error("usage calculation failed!", e)
+                CONFIG.calc_usage = False
```
stellaHSR, line 237: should add: `return usage`

metagpt/provider/openai_api.py
```diff
@@ -259,10 +263,14 @@ async def acompletion_batch_text(self, batch: list[list[dict]]) -> list[str]:
...
-            completion_tokens = int(usage['completion_tokens'])
-            self._cost_manager.update_cost(prompt_tokens, completion_tokens, self.model)
+        if CONFIG.calc_usage:
+            try:
+                prompt_tokens = int(usage['prompt_tokens'])
+                completion_tokens = int(usage['completion_tokens'])
+                self._cost_manager.update_cost(prompt_tokens, completion_tokens, self.model)
+            except Exception as e:
+                logger.error("updating costs failed!", e)
+                CONFIG.calc_usage = False
```
stellaHSR, line 273: It is better to add comments for this setting change as there will be no log information for token and money cost in case of failure.

metagpt/provider/openai_api.py
```diff
@@ -224,13 +224,17 @@ async def acompletion_text(self, messages: list[dict], stream=False) -> str:
...
 
     def _calc_usage(self, messages: list[dict], rsp: str) -> dict:
-        usage = {}
         if CONFIG.calc_usage:
-            prompt_tokens = count_message_tokens(messages, self.model)
-            completion_tokens = count_string_tokens(rsp, self.model)
-            usage['prompt_tokens'] = prompt_tokens
-            usage['completion_tokens'] = completion_tokens
-        return usage
+            usage = {}
```
stellaHSR, line 228: Please place this statement before the "if" check and then return it as default.

PR #154, opened 2023-08-08
Translate comment

martcpp:
Translate comment from native china to english to enable non chinese developer to understand what is going on


martcpp:
For easy understanding to non china developer

PR #162, opened 2023-08-08
New test added

mohammedeladwe3:


stellaHSR:
I suggest adding more unit tests, such as a unit test for the engineer prompt setting or a unit test for a specific function.

PR #190, opened 2023-08-10
Update skill_manager.py

martcpp:
Translation from Chinese to english for easy understandings by non chinese developer


martcpp:
This enable non chinese developer to understand what the comments is doing

martcpp:
These translation Still remain chinese and confusing to ko english developer

PR #199, opened 2023-08-12
feat: Add Qdrant vector search store

hezhaozhao-git:
#### commit detail

+ Add qdrant vector search for MeatGPT, about qdrant, you can check the [official website of qdrant](https://qdrant.tech/documentation/)

+ Add ruff for MetaGPT pre-commit, about ruff, you can check the [github index of ruff](https://github.com/astral-sh/ruff)

PR #207, opened 2023-08-13
add qdrant store

hezhaozhao-git:
This pr is one of https://github.com/geekan/MetaGPT/pull/199

### DIFF
--- a/metagpt/document_store/base_store.py
+++ b/metagpt/document_store/base_store.py
@@ -15,7 +15,7 @@ class BaseStore(ABC):
     """FIXME: consider add_index, set_index and think 颗粒度"""
 
     @abstractmethod
-    def search(self, query, *args, **kwargs):
+    def search(self, *args, **kwargs):
         raise NotImplementedError
 
     @abstractmethod
--- a/metagpt/document_store/qdrant_store.py
+++ b/metagpt/document_store/qdrant_store.py
@@ -10,6 +10,14 @@ from metagpt.document_store.base_store import BaseStore
 
 @dataclass
 class QdrantConnection:
+    """
+   Args:
+       url: qdrant url
+       host: qdrant host
+       port: qdrant port
+       memory: qdrant service use memory mode
+       api_key: qdrant cloud api_key
+   """
     url: str = None
     host: str = None
     port: int = None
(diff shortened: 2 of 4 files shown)


metagpt/document_store/base_store.py
```diff
@@ -15,7 +15,7 @@ class BaseStore(ABC):
     """FIXME: consider add_index, set_index and think 颗粒度"""
 
     @abstractmethod
-    def search(self, query, *args, **kwargs):
+    def search(self, *args, **kwargs):
```
better629, line 18: no need to delete, we need an explicit definition of the search `query`. And the inherited search can be `search (self, query: List[float], collection_name: str, xxx)`.

tests/metagpt/document_store/test_qdrant_store.py
```diff
@@ -0,0 +1,237 @@
...
+    qdrant_store.create_collection("Book", vectors_config, force_recreate=True)
+    assert qdrant_store.has_collection("Book") is True
+    qdrant_store.delete_collection("Book")
+    assert qdrant_store.has_collection("Book") is False
+    qdrant_store.create_collection("Book", vectors_config)
+    assert qdrant_store.has_collection("Book") is True
+    qdrant_store.add("Book", points)
+    results = qdrant_store.search("Book", query=[1.0, 1.0])
+    assert results == [
+        {
```
better629, line 47: no need check all, a few elements is proper.

metagpt/document_store/qdrant_store.py
```diff
@@ -0,0 +1,90 @@
...
+@dataclass
+class QdrantConnection:
+    url: str = None
+    host: str = None
+    port: int = None
+    memory: bool = False
+    api_key: str = None
+
+
+class QdrantStore(BaseStore):
```
better629, line 20: How to extend the usage of QdrantStore in memory, maybe you can make a supplement.

hezhaozhao-git: QdrantConnection memory is True，it will use qdrant in meory

metagpt/document_store/base_store.py
hezhaozhao-git: Although this will not report an error, it actually conflicts with the definition of the Base class. I think Base does not need to specify that there must be a query parameter, because the subclass must inherit and implement this search. As shown in the figure, pycharm will actually warn

hezhaozhao-git: This is because the method signature of the abstract class search function changed by the search method I implemented, so this warning is reported

stellaHSR:
LGTM

metagpt/document_store/qdrant_store.py
better629: The usage of QdrantStore which I mentioned is where to setup QdrantConnection in `key.yaml` or using Env methods.

PR #240, opened 2023-08-17
Make the SearchEngine more user-friendly && Fix some bugs

shenchucheng:
1. When running the SearchEngine, check if the required packages or configurations meet the requirements. If not satisfied, raise an exception and provide friendly suggestions for modification.
2. Make the researcher report output directory if it doesn't exist.
3. Fix the 'get_max_completion_tokens' function to prevent reaching the maximum limit.

By the way, the modified code file has been formatted using the pre-commit tool with the ruff and black formatting styles.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -25,12 +25,17 @@ RPM: 10
 
 #### for Search
 
+## Supported values: serpapi/google/serper/ddg
+#SEARCH_ENGINE: serpapi
+
 ## Visit https://serpapi.com/ to get key.
 #SERPAPI_API_KEY: "YOUR_API_KEY"
+
 ## Visit https://console.cloud.google.com/apis/credentials to get key.
 #GOOGLE_API_KEY: "YOUR_API_KEY"
 ## Visit https://programmablesearchengine.google.com/controlpanel/create to get id.
 #GOOGLE_CSE_ID: "YOUR_CSE_ID"
+
 ## Visit https://serper.dev/ to get key.
 #SERPER_API_KEY: "YOUR_API_KEY"
 
--- a/metagpt/config.py
+++ b/metagpt/config.py
@@ -45,8 +45,9 @@ class Config(metaclass=Singleton):
         self.global_proxy = self._get("GLOBAL_PROXY")
         self.openai_api_key = self._get("OPENAI_API_KEY")
         self.anthropic_api_key = self._get("Anthropic_API_KEY")
-        if (not self.openai_api_key or "YOUR_API_KEY" == self.openai_api_key) \
-            and (not self.anthropic_api_key or "YOUR_API_KEY" == self.anthropic_api_key):
+        if (not self.openai_api_key or "YOUR_API_KEY" == self.openai_api_key) and (
+            not self.anthropic_api_key or "YOUR_API_KEY" == self.anthropic_api_key
+        ):
             raise NotConfiguredException("Set OPENAI_API_KEY or Anthropic_API_KEY first")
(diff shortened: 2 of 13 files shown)


metagpt/config.py
```diff
@@ -45,8 +45,9 @@ def __init__(self, yaml_file=default_yaml_file):
         self.global_proxy = self._get("GLOBAL_PROXY")
         self.openai_api_key = self._get("OPENAI_API_KEY")
         self.anthropic_api_key = self._get("Anthropic_API_KEY")
-        if (not self.openai_api_key or "YOUR_API_KEY" == self.openai_api_key) \
-            and (not self.anthropic_api_key or "YOUR_API_KEY" == self.anthropic_api_key):
+        if (not self.openai_api_key or "YOUR_API_KEY" == self.openai_api_key) and (
+            not self.anthropic_api_key or "YOUR_API_KEY" == self.anthropic_api_key
```
better629, line 49: anthropic_api_key only used in `researcher` role, not in global.

metagpt/roles/researcher.py
```diff
@@ -84,10 +84,17 @@ async def _react(self) -> Message:
...
         filepath = RESEARCH_PATH / f"{topic}.md"
         filepath.write_text(content)
 
 
 if __name__ == "__main__":
-    role = Researcher(language="en-us")
-    asyncio.run(role.run("dataiku vs. datarobot"))
+    import fire
+
+    async def main(topic: str, language="en-us"):
```
better629, line 96: set default topic valie

better629:
LGTM

metagpt/roles/researcher.py
shenchucheng: Setting default values in the example is reasonable, but here, the command-line interface is provided through "fire," allowing users to directly execute python -m metagpt.roles.researcher "topic" in the terminal. If the user doesn't provide a "topic" argument, Fire will prompt them in the terminal.
```bash
python -m metagpt.roles.researcher
ERROR: The function received no value for the required argument: topic
Usage: researcher.py TOPIC <flags>
  optional flags:        --language

For detailed information on this command, run:
  researcher.py --help
```

config/config.yaml
```diff
@@ -25,12 +25,17 @@ RPM: 10
 
 #### for Search
 
+## Supported values: serpapi/google/serper/ddg
+#SEARCH_ENGINE: serpapi
```
voidking, line 29: Run with error:
```
pydantic.error_wrappers.ValidationError: 1 validation error for SerpAPIWrapper
serpapi_api_key
  To use, make sure you provide the serpapi_api_key when constructing an object. Alternatively, ensure that the environment variable SERPAPI_API_KEY is set with your API key. You can obtain an API key from https://serpapi.com/. (type=value_error)
```
We should allow users to leave this value blank.

config/config.yaml
```diff
@@ -25,12 +25,17 @@ RPM: 10
 
 #### for Search
 
+## Supported values: serpapi/google/serper/ddg
```
voidking, line 28: Should Google be used as the default search?

config/config.yaml
shenchucheng: The issue arises due to the absence of a set SERPAPI_API_KEY, which is required when using the SerpAPI search engine.

config/config.yaml
shenchucheng: In the initial version, SerpAPI was used as the default search engine. Maybe using DDG might be more suitable since it's free of charge.

voidking:
LGTM

PR #246, opened 2023-08-18
Create azure-container-webapp.yml

Antekkam:


voidking:
This pr is to realize the automatic construction and upload of the image for each commit? This is not necessary, it is better to create a mirror every time the version is stable.
Also, I found that your workflow hasn't run through yet?

geekan:

Issue #107, opened 2023-08-03
RuntimeError: Event loop is closed


liangjw:
每次执行完成后报下面错误， 是BUG还是正常

Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x00000250DE4424C0>
Traceback (most recent call last):
  File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\proactor_events.py", line 116, in __del__
    self.close()
  File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\proactor_events.py", line 108, in close
    self._loop.call_soon(self._call_connection_lost, None)
  File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\base_events.py", line 746, in call_soon
    self._check_closed()
  File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed

phhh-xh:
我也是，但是我的代码还没有执行完毕就会报这个错误

TerryFYL:
我的也是，有大佬来解答吗？

wangjie5540:
> 每次执行完成后报下面错误， 是BUG还是正常
> 
> Exception ignored in: <function _ProactorBasePipeTransport.**del** at 0x00000250DE4424C0> Traceback (most recent call last): File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\proactor_events.py", line 116, in **del** self.close() File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\proactor_events.py", line 108, in close self._loop.call_soon(self._call_connection_lost, None) File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\base_events.py", line 746, in call_soon self._check_closed() File "D:\MyData\liangjw75\AppData\Local\Programs\Python\Python39\lib\asyncio\base_events.py", line 510, in _check_closed raise RuntimeError('Event loop is closed') RuntimeError: Event loop is closed

有全一点的日志么？
个人猜测，如果与asyncio有关的，大概率是网络访问异常导致

voidking:
[...]

PR #257, opened 2023-08-22
fix(startup.py): add custom loop policy for Windows

bocandrei:
Closes: #107 
Check if the OS is windows and then set the correct asyncio policy


stellaHSR:
Great!
LGTM

PR #263, opened 2023-08-24
Adding english translation SkillManager

ImJohnDoe:
Translated method descriptions, parameters and return values using [ChatGPT](https://chat.openai.com/share/6429b7f2-0063-403d-8b45-156c5ea4a1ab). 

Translation can be checked with independent translation tool or by looking at the GPT-4 output in the above link.


stellaHSR:
has conflict, you need resolve it

Issue #167, opened 2023-08-08
Why have a dependency on npm 


romechenko:
Is there a really good reason to keep a dependency on npm? Is the package just that good?

My Take: The library is more flexible as just python

voidking:
Currently we use mermaid(nodejs) to generate image and pdf etc. If there is a better option in python, we will consider using it.

dahifi:
GPT is trained on Mermaid, and it's what others are using. What alternatives did you have in mind? 

If you don't need image/pdfs then maybe it'd be better to just have it fail gracefully, but I'd take a look at maybe doing something like this: https://mermaid.js.org/config/Tutorials.html#jupyter-integration-with-mermaid-js

ImJohnDoe:
Those seem like good reasons to use Mermaid. 

It sounds like you have two suggestions

  1. Add an option to not use Mermaid and make it fail gracefully wherever it’s used
  2. Use the [Mermaid.ink](https://mermaid.ink/) service to generate charts
  
Option 2. is nice because it solves my problem and eliminates the need for npm entirely. But it does rely on a free service that could be changed.

I like the idea of a combination of both options. Allow for disabling npm and falling back on using mermaid.ink

Any thoughts?

```
graph LR;
    A--> B & C & D;
    B--> A & E;
    C--> A & E;
    D--> A & E;
    E--> B & C & D;
```

![image](https://mermaid.ink/img/Z3JhcGggTFI7CiAgICBBLS0+IEIgJiBDICYgRDsKICAgIEItLT4gQSAmIEU7CiAgICBDLS0+IEEgJiBFOwogICAgRC0tPiBBICYgRTsKICAgIEUtLT4gQiAmIEMgJiBEOw==)

voidking:
Good idea!

ImJohnDoe:
One downside is that it does not seem to support pdf output. It does support svg, png, and jpeg.
What is the pdf format used for? Is it just to give people options?

ImJohnDoe:
Pull request opened here 
https://github.com/geekan/MetaGPT/pull/265

better629:
Since no further responses are needed, we will close it. Please reopen it if necessary.

PR #265, opened 2023-08-24
removes npm dependency by using mermaid ink service | removes referen…

romechenko:
# Remove the dependency on npm

This issue is document as [issue 167](https://github.com/geekan/MetaGPT/issues/167)

## Changes 

- removes the dependency on npm by using the mermaid ink service
- removes references to the npm version mermaid and the related issues that is causes

### Description of changes

This changes the way that mermaid.py works. Instead of doing checks to see if the npm command exists for mermaid, it makes a call to the mermaid ink service. This service takes a base64 encoded string as the mermaid code and can return either an svg or png depending on the path.

## Testing

A simple test to see if this new approach works is to run the mermaid.py file and see if the mermaid charts are created in the tmp folder in the project root. 

## Warning

This does remove the pdf file output type.

I don't see this as an issue but someone else might.

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -1,7 +1,3 @@
-# Step 1: Ensure that NPM is installed on your system. Then install mermaid-js.
-npm --version
-sudo npm install -g @mermaid-js/mermaid-cli
-
-# Step 2: Ensure that Python 3.9+ is installed on your system. You can check this by using:
+# Step 1: Ensure that Python 3.9+ is installed on your system. You can check this by using:
 python --version
 python setup.py install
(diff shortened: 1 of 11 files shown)


geekan:
This is a very good submission that can lower the threshold for users to use

But when faced with third-party network requirements, I have several questions

1. This requires a default network dependency, but this assumption may not be true in the future, such as using Llama's intranet deployment, which means it is more reasonable to make it an option?
2. What is the payment policy for third-party services? Is it free forever? It promises service stability
3. If the third-party service is down, what is our strategy?

alitrack:
had better submit as a optional solution, not directly replace the original one.
feel free to have a look mine solution https://github.com/geekan/MetaGPT/pull/289

stellaHSR:
Good idea! It may be helpful when the mermaid install fails. 
Can you resolve the conflicts? And provide test codes for this, please. 
In addition, it will be convenient if it is configurable for installation and usage

geekan:

PR #266, opened 2023-08-25
Optimize Portions of Code

isaacJinyu:


isaacJinyu:
1. Reduce code duplication: There's a lot of repeated code in `ask` and `aask`, `ask_batch` and `aask_batch`, `ask_code` and `aask_code`. Consider writing a function to handle the common parts, then call this function in each method.

2. Improve log output: In the `aask` method, the `message` is directly logged, but without providing much context information. This could complicate log inspection. Consider providing more context when logging, such as which function it's in, what message it's processing, etc.

isaacJinyu:
1. Removed `check_google_api_key` and `check_google_cse_id` validators and directly fetched environment variables in class attributes. This enhances code readability and maintainability by reducing indirection.

2. Moved default value of `focus` parameter to the function signature. This provides immediate visibility of the default value, improving code readability and usability.

3. In `safe_google_results` function, if `results` is a string, it is returned directly without encoding and decoding operations. This streamlines the code and enhances its efficiency without altering its functionality.

4. In `if __name__ == "__main__"`, an instance of `GoogleAPIWrapper` is created before calling the `fire.Fire` function. This adheres to the principles of object-oriented programming and improves the understandability and maintainability of the code.
3. Direct String Return in `safe_google_results`: This modification eliminates unnecessary string encoding and decoding operations, thereby streamlining the code and enhancing its efficiency without altering its functionality.

4. Instance Creation before `fire.Fire` Function Call: This modification adheres to the principles of object-oriented programming. It improves the understandability and maintainability of the code by creating an instance of `GoogleAPIWrapper` before calling the `fire.Fire` function, and also prepares the code for potential future expansions.

metagpt/tools/search_engine_googleapi.py
```diff
@@ -80,7 +54,7 @@ async def run(
         query: str,
         max_results: int = 8,
         as_string: bool = True,
-        focus: list[str] | None = None,
+        focus: list[str] = ["snippet", "link", "title"],
```
shenchucheng, line 57: Avoid using mutable objects as default parameters in functions.

metagpt/tools/search_engine_googleapi.py
```diff
@@ -25,38 +26,11 @@
...
 
 class GoogleAPIWrapper(BaseModel):
-    google_api_key: Optional[str] = None
-    google_cse_id: Optional[str] = None
+    google_api_key: str = os.environ.get("GOOGLE_API_KEY")
+    google_cse_id: str = os.environ.get("GOOGLE_CSE_ID")
     loop: Optional[asyncio.AbstractEventLoop] = None
     executor: Optional[futures.Executor] = None
 
-    class Config:
```
shenchucheng, line 33: This code is necessary.

metagpt/tools/search_engine_googleapi.py
```diff
@@ -25,38 +26,11 @@
 
 
 class GoogleAPIWrapper(BaseModel):
-    google_api_key: Optional[str] = None
-    google_cse_id: Optional[str] = None
+    google_api_key: str = os.environ.get("GOOGLE_API_KEY")
```
shenchucheng, line 29: It is not a good idea to get value from environ, but from CONFIG.

metagpt/tools/search_engine_googleapi.py
```diff
@@ -25,38 +26,11 @@
...
-                "To use, make sure you provide the google_api_key when constructing an object. Alternatively, "
-                "ensure that the environment variable GOOGLE_API_KEY is set with your API key. You can obtain "
-                "an API key from https://console.cloud.google.com/apis/credentials."
-            )
-        return val
-
-    @validator("google_cse_id", always=True)
-    @classmethod
-    def check_google_cse_id(cls, val: str):
-        val = val or CONFIG.google_cse_id
```
shenchucheng, line 51: When google_api_key or google_cse_id is not set, the validator will provide guidance. If delete this code and there are no google_api_key or google_cse_id, it will raise an exception, making it difficult for users to determine the cause of the error from the error message.

shenchucheng:
It appears that the command `python setup.py install` may not be able to resolve dependencies properly. You can replace it with `pip install -e .`. Alternatively, you can create a wheel file and install it using pip. For example:
```bash
python setup.py bdist_wheel
pip install dist/metagpt-0.1-py3-none-any.whl
```

metagpt/tools/search_engine_googleapi.py
isaacJinyu: Thank you for the valuable feedback on my revised code. I appreciate your insights and the time you took to review my work.

PR #276, opened 2023-08-29
feat: option to skip hiring an engineer  with implement=False

cassler:
This PR provides additional customizations on what steps to perform with `setup.py`.

- NEW: Do not hire an engineer to write code `--implement False` (default True)
- Hire an engineer and perform code review `--code_review True` (will also hire an engineer)

Additionally, there are some subjective readability improvements to the README. Glad to remove those from the diff if needed.


geekan:
LGTM

PR #291, opened 2023-09-06
Features/tutorial assistant

Stitch-z:
### Features: add a new role tutorial assistant. 
  - Function: simply input a topic to generate a Markdown formatted tutorial document based on that topic.
  - Design logic:
    - Design a prompt to obtain the directory structure of the input topic through GPT and return it according to the required structure.
    - Process the directory structure of step 1 and divide it into multiple modules based on the first level title.
    - Treat each module processed in step 2 as a write content action. Design a prompt that requires the GPT to be generated according to constraints.
    - This design can solve the length limitation problem of GPT generated text, ensuring that the content of the tutorial documents is rich and detailed.
  - The generated document effect:
 
image

image


stellaHSR:
Excellent！ Can you also provide unit test for actions ?

Stitch-z:
> Excellent！ Can you also provide unit test for actions ?

Yes, the unit tests for the role and action have been committed.

stellaHSR:
LGTM

metagpt/actions/write_tutorial.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Args:
+            topic: The tutorial topic.
+
+        Returns:
+            the tutorial directory information, such as {"title": "xxx", "directory": [{"dir 1": ["sub dir 1", "sub dir 2"]}]}
+        """
+        prompt = DIRECTORY_PROMPT.format(topic=topic, language=self.language)
+        directory = await self._aask(prompt=prompt)
+        return json.loads(directory)
```
better629, line 44: Can you handle the situation if `directory` is not a json format.

metagpt/actions/write_tutorial.py
```diff
@@ -0,0 +1,102 @@
...
+        Args:
+            title: The title of tutorial.
+            content: The total content of tutorial.
+
+        Returns:
+            The full filename of tutorial content.
+
+        """
+        current_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        pathname = TUTORIAL_PATH / current_time
```
better629, line 96: `SaveDocx` can works as a common MD saving operation, but here limit the `saving` path under `TUTORIAL_PATH`. So, the suggestion is to make it more universal or just save docx under `_react` like `https://github.com/geekan/MetaGPT/blob/main/metagpt/roles/researcher.py#L86C18-L86C18`

metagpt/prompts/tutorial_assistant.py
```diff
@@ -0,0 +1,39 @@
...
+
+COMMON_PROMPT = """
+You are now a seasoned technical professional in the field of the internet. 
+We need you to write a technical tutorial with the topic "{topic}".
+"""
+
+DIRECTORY_PROMPT = COMMON_PROMPT + """
+Please provide the specific table of contents for this tutorial, strictly following the following requirements:
+1. The output must be strictly in the specified language, {language}.
+2. Answer in the dictionary format like {{"title": "xxx", "directory": [{{"dir 1": ["sub dir 1", "sub dir 2"]}}]}}.
```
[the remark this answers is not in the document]
Stitch-z: The reason for using '{{}}' instead of '{}' in this template string is to avoid conflicts with '{}' in the template string, where "{{" indicates that it is text content "{". Otherwise, DIRECTORY_ PROMPT. format (topic="xx", language="xx") may have parsing failure issues.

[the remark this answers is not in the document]
better629: `'{lang} {{"title": "xxx"}} {topic}'.format(lang="en", topic="test", title="test2")
'en {"title": "xxx"} test'`. Do you happened with the failure case?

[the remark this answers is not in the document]
Stitch-z: image
Yes, it will throw a key error.

metagpt/actions/write_tutorial.py
Stitch-z: You are right. In rare cases, GPT will return extra opening or closing words, so I extract and format the dictionary based on '{' and '}'.

metagpt/actions/write_tutorial.py
Stitch-z: I have extracted the write file operation as a general tool method.
image

metagpt/utils/file.py
```diff
@@ -0,0 +1,42 @@
...
+
+        Raises:
+            Exception: If an unexpected error occurs during the file writing process.
+        """
+        try:
+            root_path.mkdir(parents=True, exist_ok=True)
+            full_path = root_path / filename
+            async with aiofiles.open(full_path, mode="wb") as writer:
+                await writer.write(content)
+                logger.info(f"Successfully write docx: {full_path}")
```
better629, line 38: If it's a general save func, log info should't contain `docx`

Stitch-z: Yes, it has been updated.

better629:
LGTM

PR #292, opened 2023-09-06
fix: windows compatibility

chansonzhang:
Since `command -v mmdc` won't work on windows, so we need to use "where mmdc" instead.


geekan:
Since the implemented function is very simple, it will be merged first.

PR #298, opened 2023-09-07
Feature/json write prd and design_api and project_management

femto:
fireworks.ai model is a bit stupid about formatting, so I change actions
return json to parse,
```
fireworks.ai's api is openai compatible,so we just need to change the following in key.yaml,
OPENAI_API_KEY: "xxx"
OPENAI_API_BASE: "https://api.fireworks.ai/inference/v1"
OPENAI_API_MODEL: "accounts/fireworks/models/llama-v2-13b-chat"
```

I used pre-commit in our repo, so some file changes formatting

some known problem using fireworks.ai,
1.llama-v2-13b-chat is stupid than gpt, so still from time to time generate inbalance json format,
like lacking end tag [/CONTENT], lacking ending """
2.sometimes generate code like this, which doesn't fit original WriteCode parsing format.
```
### Code: game.py

import pygame
```

so still, it generates worse than gpt


metagpt/utils/custom_decoder.py
stellaHSR, line 1: A little complex. Can you please also include the unit tests? Additionally, pls follow PEP8.

femto: added tests. fix styling thru pre-commit. sometimes llm return json single quote string,
custom_decoder is to support this since single quote string is not in json spec,
unfortunately if you want to support this, you need to touch (nearly) whole json parsing logic

metagpt/roles/qa_engineer.py
```diff
@@ -159,7 +166,7 @@ async def _act(self) -> Message:
         for msg in self._rc.news:
             # Decide what to do based on observed msg type, currently defined by human,
             # might potentially be moved to _think, that is, let the agent decides for itself
-            if msg.cause_by == WriteCode:
+            if msg.cause_by in [WriteCode, WriteCodeReview]:
```
shenchucheng, line 169: What was the reason for adding the WriteCodeReview? Perhaps it was added to fix some bugs?

requirements.txt
```diff
@@ -37,4 +37,5 @@ anthropic==0.3.6
 typing-inspect==0.8.0
 typing_extensions==4.5.0
 libcst==1.0.1
-qdrant-client==1.4.0
\ No newline at end of file
+qdrant-client==1.4.0
+pytest-mock==3.11.1
```
shenchucheng, line 41: The pytest-mock is a development dependency. Perhaps we should create a requirements-dev.txt file?

metagpt/roles/qa_engineer.py
femto: yes, fix bugs. Original code review won't trigger qa

requirements.txt
femto: then pytest is also a development dependency. What do you suggest? Perhaps another pr.
but our packaging mechanism hasn't been decided yet.

shenchucheng:
LGTM

shenchucheng (approved):
LGTM

stellaHSR:
LGTM

PR #302, opened 2023-09-09
[feat/stream-config] Enable Stream config to avoid hardcode in base o…

weburnit:
# Enable stream for completion_api
I'm working with Azure OpenAI. which does not support stream features.

Instead of having hardcode `stream=True`, I enable this config for openai_api.py.


config/config.yaml
```diff
@@ -22,7 +22,7 @@ RPM: 10
 #OPENAI_API_KEY: "YOUR_AZURE_API_KEY"
 #OPENAI_API_VERSION: "YOUR_AZURE_API_VERSION"
 #DEPLOYMENT_ID: "YOUR_DEPLOYMENT_ID"
-
+#STREAM: true
```
geekan, line 25: uncomment it would be better?

Wei-Jianan:
https://github.com/geekan/MetaGPT/issues/1257

geekan:
right. https://github.com/geekan/MetaGPT/pull/1258 solved this.

Issue #301, opened 2023-09-09
Upgrade to Pydantic 2


AByteCoder:
Hi,
Any plans on upgrading the pydantic dependency to version 2?
I see that Langchain and other libraries have started to support it.

I can start working on PR if there is no such plan.

Thanks

geekan:
Already upgrade. Actually a big change

PR #304, opened 2023-09-09
Updates requirements.txt and adds quickfix to use pydantic v2

AByteCoder:
#301 To support applications using Pydantic v2, have upgraded using a quick fix method (still using v1 internally using pydantic)
* Have Upgraded necessary dependency like Langchain, Antrophic to latest versions which support Pydantic 2.
* Have fixed the syntax error in `metagpt/tools/ut_writer.py` cause by unclosed triple quotes.
* Chromadb doesn't support pedantic v2 as of now, so it will crash until they upgrade.
* Have tested by running the startup command, building docker container and running pytest test doesn't seem to cause any additional failures in the test.

### DIFF
--- a/metagpt/actions/action_output.py
+++ b/metagpt/actions/action_output.py
@@ -8,7 +8,7 @@
 
 from typing import Dict, Type
 
-from pydantic import BaseModel, create_model, root_validator, validator
+from pydantic.v1 import BaseModel, create_model, root_validator, validator
 
 
 class ActionOutput:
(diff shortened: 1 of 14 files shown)


stellaHSR:
I fully agree with your suggestion. Utilizing callbacks can significantly enhance the agents' ability to notify other modules and the environment in a more efficient manner. They are an excellent approach for facilitating communication and collaboration among agents. 

However, after reviewing the code that you committed, I think that using hooks would be better. We can discuss this further in the pull request. Let's continue the conversation there.

PR #308, opened 2023-09-12
feature/write many code

femto:
simplifies Engineer and refactor core logic into WriteManyCodes.

the concern is, 1. the logic belongs to Action
2.when going to WriteManyCodes action, it can be triggered by other means.


geekan:

PR #312, opened 2023-09-12
Fix prompt quote bug

RainFung:
Double quotation marks not escaped before


stellaHSR:
This has been fixed, you can pull first.

PR #248, opened 2023-08-18
Update UI design actions

stellaHSR:
1. add new UI design actions:
  
- [Model Selection]: select a suitable model for a given query 

- [SDPromptOptimize]: Optimize prompt via GPT-4 to make it StableDiffusion user-friendly 

- [SDPromptImprove]: Improve prompt via GPT-4 to make it StableDiffusion friendly 

- [ SDPromptExtend]: Select prompt optimize method automatically

- [SDGeneration]: Generate images via SD T2I API

2. Update UI role think and act func:

- [plan_and_select ] After selecting the model and domain, UI can choose its method for optimizing prompts. An action is dynamically determined in the middle, which can be forked and one of two is selected to determine the appropriate one. Then this choice is executed, followed by executing the subsequent actions in order.

3. Add prompt templates and fix typo errors in sd_engine.py located in the /tools directory.

need to install json5 (pip install json5)

PR #313, opened 2023-09-12
Re-open Add UI role

stellaHSR:
https://github.com/geekan/MetaGPT/pull/248

add unit tests for resp_parse.py. All the unit tests are generated by MetaGPT!


tests/metagpt/utils/test_flatten_json_structure_json.py
```diff
@@ -0,0 +1,84 @@
...
+            items.extend(flatten_json_object(value, new_key, sep=sep).items())
+        elif isinstance(value, list):
+            items.append((new_key, sep.join(str(v) for v in value)))
+        else:
+            items.append((new_key, value))
+    return dict(items)
+
+def flatten_json_structure(json_array):
+    if (isinstance(json_array, list) and len(json_array) == 1 and not isinstance(json_array[0], str)):
+        return flatten_json_structure(json_array[0])
```
better629, line 26: the condition here do you want to deal the array like `[[{"xx", ""}]]`?

but here
>>> t1={'name': 'John', 'age': "30", 'city': 'New York'}
>>> t2=[{'name': 'John', 'age': '30', 'city': 'New York'}]
>>> t3=[{'name': 'John', 'age': '30', 'city': 'New York'}, {'k': 'g'}]
>>> flatten_json_structure(t1)
['John', '30', 'New York']
>>> flatten_json_structure(t2)
['John', '30', 'New York']
>>> flatten_json_structure(t3)
['John, 30, New York', 'g']

it seems that a dict should be flatten into 'value1, value2,..' like t3, but it got a list of t1 or t2.

tests/metagpt/utils/test_flatten_json_object.py
```diff
@@ -0,0 +1,39 @@
...
+    
+    if isinstance(obj, list):
+        return dict([("value", sep.join(str(v) for v in obj))])
+    
+    items = []
+    for key, value in obj.items():
+        new_key = f"{parent_key}{sep}{key}" if parent_key else key
+        if isinstance(value, dict):
+            items.extend(flatten_json_object(value, new_key, sep=sep).items())
+        elif isinstance(value, list):
```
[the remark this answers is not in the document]
stellaHSR: 1. yes, it is expected.
2.  update code for tuple support

tests/metagpt/utils/test_flatten_json_structure_json.py
stellaHSR: Indeed, in my case, it is used to obtain the flattened value of a dictionary.

better629: > Indeed, in my case, it is used to obtain the flattened value of a dictionary.

yes, but t3 t2 got different result with same element-0 dict. 
And `flatten_json_object` function implement and so on are repeated in `test_flatten* test_try_par*` files, maybe you should remove them and use `from .. import ` instead.

stellaHSR: Due to my prompt design, I added the original code in the test code to avoid hallucination.

better629: It's fine to add it in your auto-generate unittest stage, but there is no need to commit it.

better629:
LGTM

PR #327, opened 2023-09-16
Feature/sk agent with basic planner and action planner

femto:
```
python examples/sk_agent.py 
```

Basic Planner从会的技能挑出相关的skills组成action plan执行:
Basic Planner Example:
```
task = """
    Tomorrow is Valentine's day. I need to come up with a few date ideas. She speaks French so write it in French.
    Convert the text to uppercase"""
plan:
{
    "input": "Valentine's Day Date Ideas",
    "subtasks": [
        {"function": "WriterSkill.Brainstorm"},
        {"function": "WriterSkill.EmailTo", "args": {"recipient": "significant_other"}},
        {"function": "WriterSkill.Translate", "args": {"language": "French"}},
        {"function": "TextSkill.uppercase"}
    ]
}
```
action planner从会的技能中挑出一个skill执行task
action planner example:
```
task = "What is the sum of 110 and 990?"
```
会调出math.add执行给出1100


metagpt/planner/test_action_planner.py
```diff
@@ -0,0 +1,38 @@
...
+from semantic_kernel.planning.action_planner.action_planner import ActionPlanner
+
+from metagpt.actions import BossRequirement
+from metagpt.roles.sk_agent import SkAgent
+from metagpt.schema import Message
+
+# Get the directory of the current file
+current_file_directory = os.path.dirname(os.path.abspath(__file__))
+# Construct the skills_directory by joining the parent directory and "skillss"
+skills_directory = os.path.join(current_file_directory, "..", "skills")
```
better629, line 21: put the certain const under `metagpt/const.py`

metagpt/skills/CalendarSkill/AssistantShowCalendarEvents/config.json
```diff
@@ -0,0 +1,15 @@
+{
```
better629, line 1: suggest to add a statement in `skill/__init__.py` to ref to the `semantic-kernel/samples/skills`

metagpt/roles/sk_agent.py
```diff
@@ -0,0 +1,74 @@
...
+        super().__init__(name, profile, goal, constraints)
+        self._init_actions([ExecuteTask()])
+        self._watch([BossRequirement])
+        self.kernel = make_sk_kernel()
+
+        # how funny the interface is inconsistent
+        if planner_cls == BasicPlanner:
+            self.planner = planner_cls()
+        elif planner_cls in [SequentialPlanner, ActionPlanner]:
+            self.planner = planner_cls(self.kernel)
```
better629, line 49: add else raise(unsupported planner)

metagpt/utils/make_sk_kernel.py
```diff
@@ -0,0 +1,33 @@
...
+
+def make_sk_kernel():
+    kernel = sk.Kernel()
+    if CONFIG.openai_api_type == "azure":
+        kernel.add_chat_service(
+            "gpt-3.5", AzureChatCompletion(CONFIG.deployment_name, CONFIG.openai_api_base, CONFIG.openai_api_key)
+        )
+    else:
+        kernel.add_chat_service(
+            "gpt-3.5",
```
better629, line 27: service_id=`gpt-3.5` may cause a confuse about here only use `model gpt-3.5`, suggest to use a more common name like `chat_completion` refs to `https://learn.microsoft.com/en-us/semantic-kernel/ai-orchestration/kernel/adding-services?tabs=python#openai` or make `service_id` as a param.

metagpt/skills/CalendarSkill/AssistantShowCalendarEvents/config.json
better629: And `skills folder` seems not to be a "from ... import..." part, is it more proper to be provided in `examples/skills`?

examples/sk_agent.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+async def basic_planner_example():
+    task = """
+    Tomorrow is Valentine's day. I need to come up with a few date ideas. She speaks French so write it in French.
+    Convert the text to uppercase"""
+    role = SkAgent()
+
+    # let's give the agent some skills
+    role.import_semantic_skill_from_directory(skills_directory, "SummarizeSkill")
```
better629, line 45: For a particular role, is it proper to make these skills as default build-in ones or pass them with `SkAgent(skills=[...])`. So we can only consider the role instance and `run`.

metagpt/actions/execute_task.py
```diff
@@ -0,0 +1,17 @@
...
+@Time    : 2023/9/13 12:26
+@Author  : femto Zheng
+@File    : execute_task.py
+"""
+from metagpt.actions import Action
+from metagpt.schema import Message
+
+
+class ExecuteTask(Action):
+    def __init__(self, name="ExecuteTask", context: list[Message] = None, llm=None):
```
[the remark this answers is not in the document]
femto: I copied from WriteCode, WriteCodeReview, they all format like this.
Also I passed pre-commit hook

metagpt/planner/test_action_planner.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+@Time    : 2023/9/16 20:03
+@Author  : femto Zheng
+@File    : test_basic_planner.py
```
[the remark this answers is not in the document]
femto: changed.

metagpt/roles/sk_agent.py
femto: added, but actually I'm thinking to remove the second judgement clause,
make it like this:
```
        if planner_cls == BasicPlanner:
            self.planner = planner_cls()
        else:
            self.planner = planner_cls(self.kernel)
```
because 1. in the future, semantic kernel will add other planner
2. we will add other planner, 
so as long as the planner's constructor signature is planner_cls(self.kernel),
then it should be fine

examples/sk_agent.py
femto: for particular some specific role, we can just subclass SkAgent and hardcode its skills  in its constructor.
if you want to support skills=[...] in constructor, 
since we have both import_semantic_skill_from_directory and import_skill, these two have different signature,
what do you suggest the params look like. Also another problem is import_skill is using an object when registered,
```
role.import_skill(TextSkill(), "TextSkill")
```
let's say we instantiate some Action/Skill, that Action/Skill depend on role(like I discussed in WriteManyCodes),
```
role.import_skill(WriteManyCodes(role=self), "WriteCodeSkill")
```
then this line must happen after role's constructor

metagpt/skills/CalendarSkill/AssistantShowCalendarEvents/config.json
better629: > suggest to add a statement in `skill/__init__.py` to ref to the `semantic-kernel/samples/skills`

@femto suggest to add a statement in skill/__init__.py to ref to the semantic-kernel/samples/skills.

better629:
LGTM

stellaHSR:
I think this is a great practice, but it requires further in-depth discussion regarding the design of plans for MG's action classes and related skills. 
LGTM

PR #329, opened 2023-09-17
adding search engine option for the role of reasearch

lidaye910416:
为研究角色添加搜索引擎选项. 国内使用其他默认搜索引擎的难度比较大，例如创建serpapi，需要搞一个电话号码接验证码。

### DIFF
--- a/examples/build_customized_agent.py
+++ b/examples/build_customized_agent.py
@@ -9,6 +9,7 @@ import asyncio
 
 import fire
 
+from metagpt.llm import LLM
 from metagpt.actions import Action
 from metagpt.roles import Role
 from metagpt.schema import Message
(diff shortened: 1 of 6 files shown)


metagpt/roles/researcher.py
```diff
@@ -27,10 +28,11 @@ def __init__(
         goal: str = "Gather information and conduct research",
         constraints: str = "Ensure accuracy and relevance of information",
         language: str = "en-us",
+        engine: SearchEngineType = SearchEngineType.SERPAPI_GOOGLE,
         **kwargs,
     ):
         super().__init__(name, profile, goal, constraints, **kwargs)
-        self._init_actions([CollectLinks(name), WebBrowseAndSummarize(name), ConductResearch(name)])
+        self._init_actions([CollectLinks(name,engine), WebBrowseAndSummarize(name), ConductResearch(name)])
```
stellaHSR, line 35: PEP8

stellaHSR:
@lidaye910416 Hi, it is a good implementation for the role of researcher and related actions. 
We have an official implementation of SearchAndSummarize in /actions/search_and_summarize.py. As for CollectLinks, it is better to merge it in SearchAndSummarize. 

For instance, we can add an input parameter called "do_summarize" to indicate whether to use summarization or not.

```
class SearchAndSummarize(Action):
    def __init__(self, name="", context=None, llm=None, engine=None, search_func=None, do_summarize=False):
        self.config = Config()
        self.engine = engine or self.config.search_engine
        self.search_engine = SearchEngine(self.engine, run_func=search_func)
        self.result = ""
        self.do_summarize = do_summarize
        super().__init__(name, context, llm)
```

Regarding the run() function, if self.do_summarize is False, it can only collect links as intended, leading to a more atomized process.

shenchucheng:
> 为研究角色添加搜索引擎选项. 国内使用其他默认搜索引擎的难度比较大，例如创建serpapi，需要搞一个电话号码接验证码。

您指的应该是这个配置？
https://github.com/geekan/MetaGPT/blob/2cfa7af826767049181c6bf0b8d492ec6c0803b3/config/config.yaml#L30

目前提供了4种engine, 如果国内这四种engine都无法访问的话，应该是添加新的engine？

PR #343, opened 2023-09-19
Fix research action bug && Optimize universal file operation tools

Stitch-z:
### Fix research action bug && Optimize universal file operation tools
- Provide a universal function for extracting a list or dict from text, which may contain complex nested structures.
- Optimize file operation tools to provide block reading and writing file functions.
- Fix a bug where the research action did not process results before `json.loads()`.
95e2911a-897e-4985-9986-ffb0f0ad3648


metagpt/actions/research.py
```diff
@@ -110,7 +111,7 @@ async def run(
         system_text = system_text if system_text else RESEARCH_TOPIC_SYSTEM.format(topic=topic)
         keywords = await self._aask(SEARCH_TOPIC_PROMPT, [system_text])
         try:
-            keywords = json.loads(keywords)
+            keywords = OutputParser.extract_struct(keywords, "list")
```
geekan, line 114: should we use list instead of "list"?

metagpt/actions/write_tutorial.py
```diff
@@ -64,7 +37,7 @@ async def run(self, topic: str, *args, **kwargs) -> Dict:
         """
         prompt = DIRECTORY_PROMPT.format(topic=topic, language=self.language)
         resp = await self._aask(prompt=prompt)
-        return await self._handle_resp(resp)
+        return OutputParser.extract_struct(resp, "dict")
```
geekan, line 40: ditto

metagpt/actions/research.py
Stitch-z: Yes, it's more elegant. It has already been updated.

metagpt/utils/file.py
```diff
@@ -31,12 +34,49 @@ async def write(cls, root_path: Path, filename: str, content: bytes) -> Path:
...
+            if not file_path.exists():
+                raise FileNotFoundError(f"File not found, path is '{file_path}'")
+            chunk_size = chunk_size or cls.CHUNK_SIZE
+            async with aiofiles.open(file_path, mode="rb") as reader:
+                content = bytes()
+                while True:
+                    chunk = await reader.read(chunk_size)
+                    if not chunk:
+                        break
+                    content += chunk
```
shenchucheng, line 76: Use [bytes.join()](https://docs.python.org/3/library/stdtypes.html#bytes.join) to replace repeated concatenation.
See the note 6 in https://docs.python.org/3/library/stdtypes.html#common-sequence-operations

metagpt/utils/file.py
```diff
@@ -31,12 +34,49 @@ async def write(cls, root_path: Path, filename: str, content: bytes) -> Path:
...
+                logger.info(f"Successfully read file, the size of file: {len(content)}")
```
shenchucheng, line 77: Maybe it's a debug log.

metagpt/utils/file.py
```diff
@@ -31,12 +34,49 @@ async def write(cls, root_path: Path, filename: str, content: bytes) -> Path:
...
+            chunk_size: The size of each chunk in bytes (default is 64kb).
+
+        Returns:
+            The binary content of file.
+
+        Raises:
+            Exception: If an unexpected error occurs during the file reading process.
+        """
+        try:
+            if not file_path.exists():
```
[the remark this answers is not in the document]
Stitch-z: Yes, you are right. I just tested it and it is indeed like this.

metagpt/utils/file.py
Stitch-z: Updated.

metagpt/utils/file.py
```diff
@@ -31,12 +34,49 @@ async def write(cls, root_path: Path, filename: str, content: bytes) -> Path:
             Exception: If an unexpected error occurs during the file writing process.
         """
         try:
+            chunk_size = chunk_size or cls.CHUNK_SIZE
             root_path.mkdir(parents=True, exist_ok=True)
             full_path = root_path / filename
             async with aiofiles.open(full_path, mode="wb") as writer:
-                await writer.write(content)
+                for i in range(0, len(content), chunk_size):
```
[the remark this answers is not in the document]
Stitch-z: Yes, block writing here does not reduce memory usage, but for large files, it mainly reduces disk IO pressure.

shenchucheng (approved):
LGTM

stellaHSR:
LGTM

PR #348, opened 2023-09-21
add semantic-kernel==0.3.10.dev0 to requirements.txt

femto:
add semantic-kernel==0.3.10.dev0 to requirements.txt


shenchucheng (approved):
LGTM

PR #356, opened 2023-09-22
Changes to dependency management, build, and test running via tox

natiyazilo:
1. Dependency management via poetry to allow for easy maintenance of the vast number of inter-related dependencies
2. Bumped python to 3.10 due to some dependencies requiring this
3. Test running via `tox` command
4. Tests now run where previously some incorrect imports were preventing this (Note: They don't all pass)
5. Docker building adjusted to make use of poetry
6. Readme updated


geekan:
May I ask which dependencies require 3.10?

PR #358, opened 2023-09-24
fix ltm serialize_message

femto:
since the format has been changed to support json,
json doesn't have tuple in it.
have to use List[List[str]] instead of List[Tupe[str,str]],

while reading LongTermMemory, found out 
that code in actionoutout_schema_to_mapping still uses  List[Tupe[str,str]],
fix that

### DIFF
--- a/metagpt/utils/serialize.py
+++ b/metagpt/utils/serialize.py
@@ -4,7 +4,7 @@
 
 import copy
 import pickle
-from typing import Dict, List, Tuple
+from typing import Dict, List
 
 from metagpt.actions.action_output import ActionOutput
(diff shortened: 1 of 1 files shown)


metagpt/utils/serialize.py
```diff
@@ -38,7 +38,7 @@ def actionoutout_schema_to_mapping(schema: Dict) -> Dict:
             mapping[field] = (List[str], ...)
         elif property["type"] == "array" and property["items"]["type"] == "array":
             # here only consider the `Tuple[str, str]` situation
```
geekan, line 40: The comment should be changed in the same time

PR #363, opened 2023-09-25
Communicate with mineflayer & update on_event()

yuymf:
Simple adding：Communicate with minefalyer & update on_event() & add temp utils(need to modify)
Files in minecraft_env, it's cv from voyager/env
on_event


metagpt/utils/minecraft/json_utils.py
```diff
@@ -0,0 +1,231 @@
...
+        #     else:
+        #         # This allows the AI to react to the error message,
+        #         #   which usually results in it correcting its ways.
+        #         print("Failed to fix ai output, telling the AI.")
+        #         return json_str
+        # else:
+        raise e
+
+
+# def fix_json(json_str: str, schema: str) -> str:
```
stellaHSR, line 196: 注释的代码可以直接移除

PR #373, opened 2023-09-27
update docker image to lastest

voidking:


voidking:
From now on, the image version in the readme does not need to be modified. 
Each time building a new image, in addition to version tag, the latest version is also uploaded.

PR #398, opened 2023-10-07
Fix litellm function calling error

orange-crow:
Fix litellm function calling error and add wrapt in requirements.txt

# changed
- Comment out the statement line 44 in metagpt/tools/code_interpreter.py: interpreter.api_base = CONFIG.openai_api_base 
- add wrapt in requirements.txt


orange-crow:
Also solve this problem in #402 , so I close it.

PR #401, opened 2023-10-08
get CONFIG from ContextVar

shenchucheng:
Enable coroutine-level isolation for metagpt.Config, allowing it to integrate smoothly with asynchronous frameworks such as aiohttp, fastapi, and sanic.

### DIFF
--- a/metagpt/config.py
+++ b/metagpt/config.py
@@ -4,11 +4,13 @@
 Provide configuration, singleton
 """
 import os
+from copy import deepcopy
+from typing import Any
 
 import openai
 import yaml
 
-from metagpt.const import PROJECT_ROOT
+from metagpt.const import CONFIGS_CTX, PROJECT_ROOT
 from metagpt.logs import logger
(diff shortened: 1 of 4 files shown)


stellaHSR:
LGTM

metagpt/config.py
```diff
@@ -114,5 +120,19 @@ def get(self, key, *args, **kwargs):
             raise ValueError(f"Key '{key}' not found in environment variables or in the YAML file")
         return value
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        CONFIGS_CTX.get()[name] = value
```
better629, line 124: why not CONFIGS_CTX.set ?

shenchucheng: The value of the CONTEXTS_CTX context variable is a dictionary used to store all configuration variables, and the \_\_setattr\_\_ method is used to set new configuration variables, thereby adding them to the dictionary. CONFIGS_CTX.set is used to set a new dictionary value in the context, and this method is often called during coroutine initialization.

better629:
LTGM

Issue #326, opened 2023-09-14
源码出现了重复代码片段


dumengru:
![image](https://github.com/geekan/MetaGPT/assets/62230160/2463f4e1-7db3-40ef-bdf6-5ed072b12ac4)

dumengru:
![image](https://github.com/geekan/MetaGPT/assets/62230160/cf62cab5-32a1-4f30-9e66-cc9202e4179b)

类方法缩进....

PR #449, opened 2023-10-23
Fix issue#326 by adjust indentation

polease:


polease:
In Cost manager ,  there are three function with wrong indentation, fixed for issue #326   
  
log\_and\_reraise(retry\_state):  
get\_costs(self)    
get\_costs(self)  :

stellaHSR:
Thx

PR #462, opened 2023-10-27
feature: added configuration files  for gitpod

PentesterPriyanshu:
added .gitpod.yml  file for gitpod users this is missing so i added it


PentesterPriyanshu:
@stellaHSR  merge this pr

voidking:
I think your purpose is to automatically set up the environment in gitpod.
So there are two issues to note:
1. gitpod default python version is 3.12 which is not suitable for MetaGPT. You should change it to python3.9 or python3.10 in `.gitpod.Dockefile`
2. The init command of MetaGPT is `pip install -e.` .

PentesterPriyanshu:
> I think your purpose is to automatically set up the environment in gitpod. So there are two issues to note:
> 
>     1. gitpod default python version is 3.12 which is not suitable for MetaGPT. You should change it to python3.9 or python3.10 in `.gitpod.Dockefile`
> 
>     2. The init command of MetaGPT is `pip install -e.` .

thanks for your reply i am now commiting all changes as you want

PentesterPriyanshu:
@voidking    now you can merge it ,

geekan:

PR #483, opened 2023-11-09
update get rpm from api.openai.com

MaxZhongSheep:
I access the "https://platform.openai.com/account/limits" webpage to automatically obtain RPM by setting the session key. The logic is as follows: if the connection to OpenAI fails, the RPM is set to the default value. If the connection to OpenAI is successful but the sess-key is invalid (including when set to the api-key), an appropriate error is displayed. If the sess-key is set to "" or not set, the RPM is set to the default value. If the sess-key is valid and the connection is successful, the RPM is obtained from the webpage.


martcpp:
Provide description and pictures of you updated . For easily understanding

config/config.yaml
```diff
@@ -9,6 +9,10 @@ OPENAI_API_BASE: "https://api.openai.com/v1"
 #OPENAI_PROXY: "http://127.0.0.1:8118"
 #OPENAI_API_KEY: "YOUR_API_KEY"
 OPENAI_API_MODEL: "gpt-4"
+## If you want to automate the setup for RPM and MAX_TOKENS, please log in through your own browser at https://platform.openai.com/account/limits, enable developer mode, and find 'rate_limits' in the network section. 
+## If it's not found, refresh the recording using Ctrl+R. After locating 'rate_limits', find 'Authorization: sess-xxx' in the request header, where 'sess-xxx' is the session key.
+## If you wish to manually set RPM, please set RPM and either do not set OPENAI_LIMIT_SESSION_KEY or set it to "".
+# OPENAI_LIMIT_SESSION_KEY: "YOUR_API_LIMITS_KEY"
```
better629, line 15: OPENAI_LIMIT_SESSION_KEY -> OPENAI_SESSION_KEY
session_key is a general key, not only for `limit`

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,29 @@ async def amoderation(self, content: Union[str, list[str]]):
...
+                                        timeout=10,
+                                        proxies={'https': openai.proxy}
+                                        )
+            except Exception as e:
+                logger.error(f"Connection to api.openai.com failed:{e}.Setting rpm to default parameter.")
+                return default_rpm
+            
+            if response.status_code == 200:
+                limit_dict = json.loads(response.text)[config.openai_api_model]
+                return limit_dict['max_requests_per_1_minute']
```
better629, line 347: different model has different rpm, not just a value

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,29 @@ async def amoderation(self, content: Union[str, list[str]]):
...
+            else:
+                error = json.loads(response.text)['error']
+                raise ValueError(f"Get rpm from api.openai.com error. {error['message']}")
```
better629, line 350: why not just use default rpm value to make the process continue

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,29 @@ async def amoderation(self, content: Union[str, list[str]]):
     async def _amoderation(self, content: Union[str, list[str]]):
         rsp = await self.llm.Moderation.acreate(input=content)
         return rsp
+
+    @staticmethod
+    def __get_rpm(config) -> int:
```
better629, line 330: why __get_rpm not _get_rpm
missing input param type

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,29 @@ async def amoderation(self, content: Union[str, list[str]]):
     async def _amoderation(self, content: Union[str, list[str]]):
         rsp = await self.llm.Moderation.acreate(input=content)
         return rsp
+
+    @staticmethod
```
better629, line 329: why staticmethod?

metagpt/provider/openai_api.py
MaxZhongSheep: > different model has different rpm, not just a value
I have indexed response.text by config.openai_api_model

metagpt/provider/openai_api.py
MaxZhongSheep: > why staticmethod?

Because it does not use any members of self.

metagpt/provider/openai_api.py
MaxZhongSheep: > why __get_rpm not _get_rpm missing input param type

I use __get_rpm instead of _get_rpm because this method should not be called in subclasses, and it can be accessed through self.rpm alone.

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,28 @@ async def amoderation(self, content: Union[str, list[str]]):
...
+                                        )
+            except Exception as e:
+                logger.error(f"Connection to api.openai.com failed:{e}.Setting rpm to default parameter.")
+                return default_rpm
+            
+            if response.status_code == 200:
+                limit_dict = json.loads(response.text)[config.openai_api_model]
+                return limit_dict['max_requests_per_1_minute']
+            else:
+                error = json.loads(response.text)['error']
```
better629, line 349: contains ' and ", just use "

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,28 @@ async def amoderation(self, content: Union[str, list[str]]):
...
+                                        headers={'Authorization': f'Bearer {session_key}'},
+                                        timeout=10,
+                                        proxies={'https': openai.proxy}
+                                        )
+            except Exception as e:
+                logger.error(f"Connection to api.openai.com failed:{e}.Setting rpm to default parameter.")
+                return default_rpm
+            
+            if response.status_code == 200:
+                limit_dict = json.loads(response.text)[config.openai_api_model]
```
better629, line 346: what if the user configured openai_api_model value not in the response's json

metagpt/provider/openai_api.py
MaxZhongSheep: > different model has different rpm, not just a value

I have indexed response.text by config.openai_api_model

metagpt/provider/openai_api.py
MaxZhongSheep: > what if the user configured openai_api_model value not in the response's json

This response's json contains all of the models supported by openai. So if the user put openai_api_model value which not in keys of response, it means that user put a wrong model name and it should raise a exception.

metagpt/provider/openai_api.py
```diff
@@ -323,3 +325,28 @@ async def amoderation(self, content: Union[str, list[str]]):
...
         rsp = await self.llm.Moderation.acreate(input=content)
         return rsp
+
+    @staticmethod
+    def __get_rpm(config: Config) -> int:
+        session_key = config.get("OPENAI_SESSION_KEY", "")
+        default_rpm = int(config.get("RPM", 10))
+        if len(session_key) > 0:
+            try:
+                response = requests.get(
```
shenchucheng, line 335: Consider using `aiohttp` instead of `requests` for asynchronous and non-blocking HTTP requests.

metagpt/provider/openai_api.py
MaxZhongSheep: This was replace by _get_rpm

metagpt/provider/openai_api.py
```diff
@@ -33,16 +36,57 @@
...
         return [batch[i : i + self.rpm] for i in range(0, len(batch), self.rpm)]
 
+    async def update_rpm(self):
+        if self.rpm is None:
+            self.rpm = await self._aget_rpm()
+            self.interval = 1.1 * 60 / self.rpm
+            logger.info(f'Setting rpm to {self.rpm}')
+
+    async def _aget_rpm(self):
+        session_key = CONFIG.get("OPENAI_SESSION_KEY", "")
```
stellaHSR, line 58: 适当添加下函数说明，以及返回结果数据类型和说明

stellaHSR:
提供下对应的测试用例脚本或者examples

geekan:

tests/metagpt/provider/test_openai_api.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+
+from metagpt.logs import logger
+from metagpt.provider.openai_api import OpenAILLM
+from metagpt.config2 import config
+
+@pytest.mark.asyncio
+async def test_update_rpm():
+    llm = OpenAILLM(config)
```
geekan, line 17: It accepts LLMConfig, not Config

geekan:

Issue #477, opened 2023-11-04
Exceeded token limit - Contacting the developer of metagpt


gnovelli:
[error.txt](https://github.com/geekan/MetaGPT/files/13256111/error.txt)

It looks like the code encountered an error due to exceeding the token limit set by the model. This happened when attempting to generate a completion using the OpenAI API, specifically for the model that has a maximum context length of 8192 tokens. However, the request included 8298 tokens, which is beyond the allowed limit.

The error trace is showing that the problem occurred in an asynchronous call within a Python script that is part of a larger application, perhaps one that is simulating a software company environment where an "engineer" role is trying to generate code through a `WriteCode` action.

Here's a breakdown of the issue:

1. The `write_code` method in `write_code.py` is called to generate some code.
2. This method makes an asynchronous request to the OpenAI API to get a response.
3. The request fails because the combined length of the messages and the expected completion exceeds the model's token limit.
4. Tenacity, a library for retrying failed operations, attempts to retry the operation, but it keeps failing because the token limit issue is not resolved.
5. Eventually, Tenacity stops retrying and raises a `RetryError`, which is not caught, causing the whole application to crash.

To fix this issue, you would need to:

- Reduce the size of the input prompt or the expected length of the completion so that the total number of tokens does not exceed the model's limit of 8192 tokens.
- Catch and handle the `InvalidRequestError` exception to ensure the application does not crash, possibly by adjusting the input and retrying.
- If the input cannot be reduced, consider breaking down the task into smaller parts that can be processed within the token limit and then combine the results. 

If you're looking to adjust the script, it would involve catching the exception where the `write_code` method is called and then trimming the input. If you have access to the code, you can modify the relevant part of the script to handle this case. **Otherwise, you may need to contact the developer of the application if you're not able to modify the code yourself.**

martcpp:
Will modify this today check back Later thanks

Issue #255, opened 2023-08-22
Rate limit error


gnovelli:
2023-08-22 07:53:00.873 | INFO     | metagpt.actions.write_code:run:77 - Writing models.py..
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/tenacity/_asyncio.py", line 50, in __call__
    result = await fn(*args, **kwargs)
  File "/app/metagpt/metagpt/actions/write_code.py", line 71, in write_code
    code_rsp = await self._aask(prompt)
  File "/app/metagpt/metagpt/actions/action.py", line 47, in _aask
    return await self.llm.aask(prompt, system_msgs)
  File "/app/metagpt/metagpt/provider/base_gpt_api.py", line 44, in aask
    rsp = await self.acompletion_text(message, stream=True)
  File "/app/metagpt/metagpt/provider/openai_api.py", line 32, in wrapper
    return await f(*args, **kwargs)
  File "/app/metagpt/metagpt/provider/openai_api.py", line 218, in acompletion_text
    return await self._achat_completion_stream(messages)
  File "/app/metagpt/metagpt/provider/openai_api.py", line 151, in _achat_completion_stream
    response = await openai.ChatCompletion.acreate(
  File "/usr/local/lib/python3.9/site-packages/openai/api_resources/chat_completion.py", line 45, in acreate
    return await super().acreate(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/openai/api_resources/abstract/engine_api_resource.py", line 217, in acreate
    response, _, api_key = await requestor.arequest(
  File "/usr/local/lib/python3.9/site-packages/openai/api_requestor.py", line 382, in arequest
    resp, got_stream = await self._interpret_async_response(result, stream)
  File "/usr/local/lib/python3.9/site-packages/openai/api_requestor.py", line 726, in _interpret_async_response
    self._interpret_response_line(
  File "/usr/local/lib/python3.9/site-packages/openai/api_requestor.py", line 763, in _interpret_response_line
    raise self.handle_error_response(
openai.error.RateLimitError: Rate limit reached for 10KTPM-200RPM in organization org-tzt146yc7FwoDKlyH63f2AH7 on tokens per min. Limit: 10000 / min. Please try again in 6ms. Contact us through our help center at help.openai.com if you continue to have issues.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/metagpt/startup.py", line 40, in <module>
[...]

Issue #253, opened 2023-08-21
openai.error.RateLimitError: Rate limit reached for 10KTPM-200RPM in organization org-mSnj0BKpxchO5RMRJ0Q7zro0 on tokens per min. Limit: 10000 / min. Please try again in 6ms.


gphett:
need to insert a pause so we do not get this error

gphett:
async def _achat_completion_stream(self, messages: list[dict]) -> str:
        try:
            response = await openai.ChatCompletion.acreate(**self._cons_kwargs(messages), stream=True)  # Using acreate if it's async

            # create variables to collect the stream of chunks
            collected_chunks = []
            collected_messages = []
            # iterate through the stream of events
            async for chunk in response:
                collected_chunks.append(chunk)  # save the event response
                chunk_message = chunk["choices"][0]["delta"]  # extract the message
                collected_messages.append(chunk_message)  # save the message
                if "content" in chunk_message:
                    print(chunk_message["content"], end="")
            print()

            full_reply_content = "".join([m.get("content", "") for m in collected_messages])
            usage = self._calc_usage(messages, full_reply_content)
            self._update_costs(usage)
            return full_reply_content
        except RateLimitError as e:  # Correctly imported
            logger.error(f"Rate limit error: {e}")
            raise  # Re-raise the error for the upper layers to handle


This seems to have fixed it for me

dvizcaya91:
I think tenacity is supposed to be the one that handles the retries, but this is so far the only solution that worked for me. Thanks

pablospe:
Perhaps you could provide a Pull Request?

geekan:
check https://github.com/geekan/MetaGPT/blob/main/config/config.yaml RPM para

thatiodeguy:
T

> from openai.error import RateLimitError
> 
> async def _achat_completion_stream(self, messages: list[dict]) -> str: retry_delay = 60 # Time in seconds to wait before retrying max_retries = 10 # Maximum number of retries
> 
> ```
> for attempt in range(max_retries):
>     try:
>         response = await openai.ChatCompletion.acreate(**self._cons_kwargs(messages), stream=True)
> 
>         # Create variables to collect the stream of chunks
>         collected_chunks = []
>         collected_messages = []
>         # Iterate through the stream of events
>         async for chunk in response:
>             collected_chunks.append(chunk)  # Save the event response
>             chunk_message = chunk["choices"][0]["delta"]  # Extract the message
[...]

PR #485, opened 2023-11-12
replaced wait_fixed with wait_exponential

a-cid:
Fixed MetaGPT crashing due to `openai.error.RateLimitError`.

The best way to handle API rate limits is by using an exponential backoff retry strategy. This is done by simply replacing tenacity's `wait_fixed` with `wait_exponential`, and to be extra sure, increasing the number of attempts by one.

Related open issues:

https://github.com/geekan/MetaGPT/issues/253
https://github.com/geekan/MetaGPT/issues/255
https://github.com/geekan/MetaGPT/issues/477

### DIFF
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -9,7 +9,7 @@ import re
 from abc import ABC
 from typing import Optional
 
-from tenacity import retry, stop_after_attempt, wait_fixed
+from tenacity import retry, stop_after_attempt, wait_exponential
 
 from metagpt.actions.action_output import ActionOutput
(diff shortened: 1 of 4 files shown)


a-cid:
> Why not use "stop=stop_after_attempt(6), wait=wait_random_exponential(min=1, max=60)" (according to [OpenAI Cookbook](https://cookbook.openai.com/examples/how_to_handle_rate_limits)), have you tried it?

Thanks for the reference! I've changed the retries to follow the cookbook's suggestion.

stellaHSR:
LGTM

PR #486, opened 2023-11-13
Add customized LLM function

YangQianli92:


config/Customized.py
```diff
@@ -0,0 +1,108 @@
+# Configurations for customizing LLM
```
better629, line 1: use lower case filename. and config file suggest to use yaml

config/Customized.py
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+ACTION_LLMS_MAPPING = {
+
+    "BossRequirement":None,
+
+    "WriteCode":{
+        'llm':{
+            "api":"https://api.openai.com/v1",
```
better629, line 61: default llm not need to write 2+ times, you can use the LLM in key.yaml

config/config.yaml
```diff
@@ -5,10 +5,11 @@
 ## The official OPENAI_API_BASE is https://api.openai.com/v1
 ## If the official OPENAI_API_BASE is not available, we recommend using the [openai-forward](https://github.com/beidongjiedeguang/openai-forward).
 ## Or, you can configure OPENAI_PROXY to access official OPENAI_API_BASE.
-OPENAI_API_BASE: "https://api.openai.com/v1"
+#OPENAI_API_BASE: "https://api.openai.com/v1"
 #OPENAI_PROXY: "http://127.0.0.1:8118"
-#OPENAI_API_KEY: "YOUR_API_KEY"
-OPENAI_API_MODEL: "gpt-4"
+OPENAI_API_BASE : ""
```
better629, line 10: don't commit non-new content if you just update the config value

metagpt/actions/action.py
```diff
@@ -47,9 +46,12 @@ async def _aask(self, prompt: str, system_msgs: Optional[list[str]] = None) -> s
         if not system_msgs:
             system_msgs = []
         system_msgs.append(self.prefix)
+
+        self.llm._CustomedGPTAPI__init_openai()
```
better629, line 50: keep uniform lower-case

metagpt/actions/design_api.py
```diff
@@ -151,6 +151,7 @@ class Game{
 
 class WriteDesign(Action):
     def __init__(self, name, context=None, llm=None):
+        llm = ACTION_LLMS_MAPPING['WriteDesign']
```
better629, line 154: Here llm is a dict but not a LLM instance, seems incorrect

metagpt/provider/openai_api.py
```diff
@@ -323,3 +329,40 @@ async def amoderation(self, content: Union[str, list[str]]):
     async def _amoderation(self, content: Union[str, list[str]]):
         rsp = await self.llm.Moderation.acreate(input=content)
         return rsp
+
+
+class CustomedGPTAPI(OpenAIGPTAPI):
```
better629, line 334: Should inherited from BaseGPTAPI, because custom api can be wenxingyiyan or spark api.
But if you means that it's the models which keep the same openai api format, it's fine.

metagpt/provider/openai_api.py
```diff
@@ -323,3 +329,40 @@ async def amoderation(self, content: Union[str, list[str]]):
...
+        self.customedllm = customedllm
+
+        self.llm = openai
+        self.model = CONFIG.openai_api_model if customedllm is None else customedllm['llm']['model_name']
+        #self.__init_openai(CONFIG)
+        #self.model = CONFIG.openai_api_model
+        self.config = CONFIG
+        self.auto_max_tokens = False
+        self._cost_manager = CostManager()
+        self.rpm = int(self.config.get("RPM", 10))
```
better629, line 353: No need to inherit again if you have parent class

metagpt/actions/action.py
```diff
@@ -16,14 +16,13 @@
...
+from metagpt.provider.openai_api import CustomedGPTAPI as CustomedLLM
 
 class Action(ABC):
     def __init__(self, name: str = "", context=None, llm: LLM = None):
         self.name: str = name
-        if llm is None:
-            llm = LLM()
-        self.llm = llm
+
+        self.llm = CustomedLLM(llm)
```
better629, line 25: Suggest to init the LLM according  to the configuration.
1. If it's openai model config, use OpenAIGPTAPI
2. If it's custom model but with openai api format, you can use OpenAIGPTAPI also
3. If it's third-part llm like claude, you can use Claude2
In this order, the use only consider the role/action's llm config.

metagpt/roles/role.py
```diff
@@ -93,8 +93,8 @@ def history(self) -> list[Message]:
 class Role:
     """Role/Agent"""
 
-    def __init__(self, name="", profile="", goal="", constraints="", desc=""):
-        self._llm = LLM()
+    def __init__(self, name="", profile="", goal="", constraints="", desc="",customedllm = None):
```
better629, line 96: add space after `,` and add input param type

metagpt/roles/role.py
```diff
@@ -109,7 +109,7 @@ def _init_actions(self, actions):
         self._reset()
         for idx, action in enumerate(actions):
             if not isinstance(action, Action):
-                i = action("")
+                i = action("")
```
better629, line 112: remove useless space

metagpt/roles/role.py
```diff
@@ -152,6 +152,9 @@ async def _think(self) -> None:
         prompt = self._get_prefix()
         prompt += STATE_TEMPLATE.format(history=self._rc.history, states="\n".join(self._states),
                                         n_states=len(self._states) - 1)
+
+        self._llm._CustomedGPTAPI__init_openai()
```
better629, line 156: It seems that `_init_openai` has been called in `__init__`, why call it again.

PR #500, opened 2023-11-22
Feat repair llmoutput

better629:
Feat
1. Try to repair the llm raw output when using opensource llm due to the weak instruction following ability
2. Add a tag of REPAIR_LLM_OUTPUT in config.yaml. If it's false, it will run the original pipeline
3. Most repair cases are in the `tests/metagpt/utils/test_repair_llm_raw_output.py`, you can have a check.
4. Add a ahttp_client to simplify the usage of aiohttp

Notice
1. It's an experimental feature. If you turn on the switch, it means that you want to have a try on running MetaGPT with open llm which may not get a ideal result. But we are working on improving it.

open llm  log(log info may be updated, but keywords exists)
[repair_llm_raw_output.log](https://github.com/geekan/MetaGPT/files/13433792/repair_llm_raw_output.log)

To_Improve
1. the regex pattern of extract content from [CONTENT]xx[/CONTENT] may should be improved.

### DIFF
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -5,17 +5,16 @@
 @Author  : alexanderwu
 @File    : action.py
 """
-import re
+
 from abc import ABC
 from typing import Optional
 
-from tenacity import retry, stop_after_attempt, wait_fixed
+from tenacity import retry, stop_after_attempt, wait_fixed, after_log
 
 from metagpt.actions.action_output import ActionOutput
 from metagpt.llm import LLM
 from metagpt.logs import logger
 from metagpt.utils.common import OutputParser
-from metagpt.utils.custom_decoder import CustomDecoder
 from metagpt.utils.repair_llm_raw_output import repair_llm_raw_output, RepairType,\
     retry_parse_json_text, extract_content_from_output
 
@@ -51,7 +50,11 @@ class Action(ABC):
         system_msgs.append(self.prefix)
         return await self.llm.aask(prompt, system_msgs)
 
-    # @retry(stop=stop_after_attempt(3), wait=wait_fixed(1))
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(1),
+        after=after_log(logger, logger.level("ERROR").name),
+    )
     async def _aask_v1(
         self,
         prompt: str,
@@ -65,7 +68,7 @@ class Action(ABC):
             system_msgs = []
         system_msgs.append(self.prefix)
         content = await self.llm.aask(prompt, system_msgs)
-        logger.debug(content)
+        logger.debug(f"llm raw output:\n{content}")
         output_class = ActionOutput.create_model_class(output_class_name, output_data_mapping)
         output_class_fields = list(output_class.schema()["properties"].keys())  # Custom ActionOutput's fields
 
@@ -73,8 +76,8 @@ class Action(ABC):
             content = repair_llm_raw_output(content, req_keys=output_class_fields + ["[/CONTENT]"])
             content = extract_content_from_output(content)
             content = repair_llm_raw_output(content, req_keys=[None], repair_type=RepairType.JSON)  # req_keys mocked
-            logger.info(f"extracted CONTENT from content:\n{content}")
-            parsed_data = retry_parse_json_text(content)
+            logger.info(f"extracted json CONTENT from output:\n{content}")
+            parsed_data = retry_parse_json_text(output=content)  # should use output=content
 
         else:  # using markdown parser
             parsed_data = OutputParser.parse_data_with_mapping(content, output_data_mapping)
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -6,7 +6,6 @@
 @File    : action.py
 """
 
-import typing
 from abc import ABC
 from typing import Optional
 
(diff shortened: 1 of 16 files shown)


metagpt/utils/repair_llm_raw_output.py
```diff
@@ -0,0 +1,300 @@
...
+    logger.debug(f"output to json decode:\n{output}")
+
+    # if CONFIG.repair_llm_output is True, it will try to fix output until the retry break
+    parsed_data = CustomDecoder(strict=False).decode(output)
+
+    return parsed_data
+
+
+def extract_content_from_output(content: str, right_key: str = "[/CONTENT]"):
+    """ extract xxx from [CONTENT](xxx)[/CONTENT] using regex pattern """
```
better629, line 259: @shenchucheng you can have a check about the extraction pattern.
The common cases are list in the `test_repair_llm_raw_output.py`, there is one case can't pass with original extract pattern, so I change it. But I think it's not good enough.

Target: find a regex pattern can handle most cases of generated output.

metagpt/actions/action.py
```diff
@@ -63,24 +68,21 @@ async def _aask_v1(
             system_msgs = []
         system_msgs.append(self.prefix)
         content = await self.llm.aask(prompt, system_msgs)
-        logger.debug(content)
+        logger.debug(f"llm raw output:\n{content}")
         output_class = ActionOutput.create_model_class(output_class_name, output_data_mapping)
+        output_class_fields = list(output_class.schema()["properties"].keys())  # Custom ActionOutput's fields
 
         if format == "json":
```
stellaHSR, line 75: do we need to use it when llm is GPT-4?

metagpt/roles/role.py
```diff
@@ -79,11 +83,11 @@ class RoleContext(BaseModel):
...
-    state: int = Field(default=-1) # -1 indicates initial or termination state where todo is None
+    state: int = Field(default=-1)  # -1 indicates initial or termination state where todo is None
     todo: Action = Field(default=None)
     watch: set[Type[Action]] = Field(default_factory=set)
     news: list[Type[Message]] = Field(default=[])
-    react_mode: RoleReactMode = RoleReactMode.REACT # see `Role._set_react_mode` for definitions of the following two attributes
+    react_mode: RoleReactMode = RoleReactMode.REACT  # see `Role._set_react_mode` for definitions of the following two attributes
     max_react_loop: int = 1
 
     class Config:
```
stellaHSR, line 93: This may be ambiguous. How about RoleMemoryConfig or a different Class name?

stellaHSR, line 93: This may be ambiguous. How about RoleMemoryConfig or a different Class name?

metagpt/actions/action.py
better629: gpt-4 allways gets right instruction result, REPAIR_LLM_OUTPUT can be False

metagpt/roles/role.py
better629: not provide by me. and it seems a pydantic built-in method.

metagpt/utils/repair_llm_raw_output.py
```diff
@@ -0,0 +1,308 @@
...
+            right_key = req_key
+        else:
+            left_key = req_key
+            right_key = f"{req_key[0]}{sc}{req_key[1:]}"  # `[req_key]` -> `[/req_key]`
+
+        if left_key not in output:
+            output = left_key + "\n" + output
+        if right_key not in output:
+            def judge_potential_json(routput: str, left_key: str) -> Union[str]:
+                routput = copy.deepcopy(routput)
```
shenchucheng, line 86: Performing a deep copy on `routput` is likely unnecessary as routput is a string, which is an immutable object.

metagpt/provider/open_llm_api.py
```diff
@@ -0,0 +1,47 @@
...
+            f"prompt_tokens: {prompt_tokens}, completion_tokens: {completion_tokens}"
+        )
+        CONFIG.total_cost = self.total_cost
+
+
+class OpenLLMGPTAPI(OpenAIGPTAPI):
+
+    def __init__(self):
+        self.__init_openllm(CONFIG)
+        self.llm = openai
```
iorisa, line 38: openai 1.x版本这种用法被废弃了

metagpt/actions/action.py
```diff
@@ -49,7 +50,11 @@ async def _aask(self, prompt: str, system_msgs: Optional[list[str]] = None) -> s
         system_msgs.append(self.prefix)
         return await self.llm.aask(prompt, system_msgs)
 
-    @retry(stop=stop_after_attempt(3), wait=wait_fixed(1))
+    @retry(
```
seehi, line 53: main代码已经使用指数回避重试，代码应该有冲突

metagpt/utils/utils.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python
```
seehi, line 1: utils目录下有个common.py，如果新建utils.py，两者的定位区别是？

metagpt/provider/fireworks_api.py
```diff
@@ -0,0 +1,24 @@
...
+
+from metagpt.config import CONFIG
+from metagpt.provider.openai_api import OpenAIGPTAPI, CostManager, RateLimiter
+
+
+class FireWorksGPTAPI(OpenAIGPTAPI):
+
+    def __init__(self):
+        self.__init_fireworks(CONFIG)
+        self.llm = openai
```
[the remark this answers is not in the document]
better629: will merge the updated openai sdk code later

metagpt/utils/utils.py
better629: can merged with common.py if all these are tool functions

seehi:
LGTM

PR #525, opened 2023-12-04
fixbug: RFC 135

iorisa:
feat: +SummarizeCode
feat: Move the logic related to project_name to WritePRD.
feat: Add unit tests
feat: The log for incrementally generated code can be referred to in the appendix section of [RFC 135](https://deepwisdom.feishu.cn/wiki/B1bcwZRdPiC8OYkweZYc9qWinSb).

PR #531, opened 2023-12-05
fixbug: Unclosed connection caused by openai 0.x. Upgrade openai to 1.x, DEP #525

iorisa:
fixbug: [Unclosed connection caused by openai 0.x](https://sentry.deepwisdom.ai/share/issue/87850db8701540e6a7b055fd50f766df/). Upgrade openai to 1.x. DEP [#525](https://github.com/geekan/MetaGPT/pull/525/files)

1. `BaseGPTAPI` class的所有接口增加timeout参数
2.  upgrade semantic-kernel
3. upgrade websocket-client==1.6.2

### DIFF
--- a/metagpt/actions/design_api.py
+++ b/metagpt/actions/design_api.py
@@ -11,9 +11,9 @@
 """
 import json
 from pathlib import Path
-from typing import List
 
 from metagpt.actions import Action, ActionOutput
+from metagpt.actions.design_api_an import DESIGN_API_NODE
 from metagpt.config import CONFIG
 from metagpt.const import (
     DATA_API_DESIGN_FILE_REPO,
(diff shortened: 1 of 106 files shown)


metagpt/actions/design_api.py
```diff
@@ -229,50 +229,21 @@ async def run(self, with_messages, format=CONFIG.prompt_format):
 
     async def _new_system_design(self, context, format=CONFIG.prompt_format):
         prompt_template, format_example = get_template(templates, format)
+        format_example = format_example.format(project_name=CONFIG.project_name)
```
weburnit, line 232: I think AGENT will seed the project name from the prompt-context rather than config.

metagpt/actions/design_api.py
```diff
@@ -58,15 +58,15 @@
 """,
         "FORMAT_EXAMPLE": """
 [CONTENT]
-{
+{{
     "Implementation approach": "We will ...",
-    "project_name": "snake_game",
+    "Project name": "{project_name}",
```
weburnit, line 63: snake_game contribute this prompt as an example blog. we shouldn't populate context data

metagpt/actions/design_api.py
```diff
@@ -58,15 +58,15 @@
...
     "File list": ["main.py"],
     "Data structures and interfaces": '
     classDiagram
-        class Game{
```
weburnit, line 67: This original format follows UML, I dont think `{{` makes it better for this prompt

metagpt/actions/design_api.py
iorisa: That feature is still present; now, an additional function has been added to allow users to specify a project name.

metagpt/actions/design_api.py
```diff
@@ -173,7 +173,7 @@ class Game{
 
 ## Implementation approach: Provide as Plain text. Analyze the difficult points of the requirements, select the appropriate open-source framework.
 
-## project_name: Provide as Plain text, concise and clear, characters only use a combination of all lowercase and underscores
+## Project name: Constant text "{project_name}".
```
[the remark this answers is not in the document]
iorisa: If the user does not set a project name, the process will generate a project name during the `WritePRD` phase. Therefore, during the design stage, the project name is already a valid string constant.

metagpt/actions/design_api.py
iorisa: During the design stage, the project name is already a valid string constant.

metagpt/actions/design_api.py
iorisa: `{{` is used in string formatting, and the result returned by the format operation is `{`.

PR #543, opened 2023-12-09
[CHORE] Research Role allow override sys prompt

weburnit:
# Overview
I think each action serve one particular role always has a chance to override its system prompt.

For instance: CollectLinks doesn't need heavy prompt, but ConductResearch is able to inherit something from parent role


shenchucheng (approved):
LGTM

stellaHSR:
LGTM

PR #553, opened 2023-12-12
fixbug: WriteCode adds the content of the related code files.

iorisa:
fixbug: WriteCode adds the content of the related code files.


geekan:
LGTM

PR #555, opened 2023-12-12
feat: After users provide bug feedback, move directly to the WriteCode stage of the process.

iorisa:
feat: After users provide bug feedback, move directly to the WriteCode stage of the process.

### DIFF
--- a/metagpt/utils/git_repository.py
+++ b/metagpt/utils/git_repository.py
@@ -8,15 +8,13 @@
 """
 from __future__ import annotations
 
-import os
+from gitignore_parser import parse_gitignore, rule_from_pattern, handle_negation
 import shutil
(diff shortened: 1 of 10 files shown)


metagpt/utils/git_repository.py
```diff
@@ -244,6 +273,7 @@ def get_files(self, relative_path: Path | str, root_relative_path: Path | str =
 
     repo = GitRepository()
     repo.open(path, auto_init=True)
+    repo.filter_gitignore(filenames=["snake_game/snake_game/__pycache__", "snake_game/snake_game/game.py"])
```
geekan, line 276: 这个的作用是？

geekan:
LGTM

PR #572, opened 2023-12-16
pref: optimize logs

Undertone0809:
Increase log rotation, allowing for automatic daily creation of daily logs under long-term operation.


geekan:
Now there is a function of rotating by day. There will be no actual functional difference between adding these lines of code and modifying it before.

Issue #629, opened 2023-12-25
Communicating error when using ZHIPUAI as the large model provider in an HTTP proxy environment


springasa:
**Bug description**
When using ZHIPUAI as the large model provider, it is not possible to access ZHIPUAI in an HTTP proxy environment, and the following error will be reported if you run build_customized_agent.py:
 ```
openai.error.APIConnectionError: Error communicating with OpenAI
 ```

**Bug solved method**

Set proxy for class ZhiPuAPTAPI. 
See the Pull Request for details: https://github.com/geekan/MetaGPT/pull/630

**Environment information**

- LLM type and model name: chatglm_turbo
- System version: 
- Python version:   Python 3.11.4

- packages version:
- installation method: 

**Screenshots or logs**
 ```
Traceback (most recent call last):
  File "D:\workspace\code\MetaGPT\venv\Lib\site-packages\aiohttp\connector.py", line 980, in _wrap_create_connection
    return await self._loop.create_connection(*args, **kwargs)  # type: ignore[return-value]  # noqa
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\springasa\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 1085, in create_connection
    raise exceptions[0]
  File "C:\Users\springasa\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 1069, in create_connection
    sock = await self._connect_sock(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\springasa\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 973, in _connect_sock
    await self.sock_connect(sock, address)
  File "C:\Users\springasa\AppData\Local\Programs\Python\Python311\Lib\asyncio\proactor_events.py", line 726, in sock_connect
    return await self._proactor.connect(sock, address)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\springasa\AppData\Local\Programs\Python\Python311\Lib\asyncio\windows_events.py", line 846, in _poll
    value = callback(transferred, key, ov)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #630, opened 2023-12-25
[Bugfix] Set openai proxy for class ZhiPuAPTAPI

springasa:
When using ZHIPUAI as the large model provider, it is not possible to access ZHIPUAI in an HTTP proxy environment, and the following error will be reported: openai.error.APIConnectionError: Error communicating with OpenAI

So we need set proxy for class ZhiPuAPTAPI.

**Features**

https://github.com/geekan/MetaGPT/issues/629

**Feature Docs**

**Influence**

**Result**
 ```
@springasa ➜ /workspaces/MetaGPT (bug-fix-proxy) $ pytest -vs tests/metagpt/provider/test_zhipuai_api.py 
2023-12-25 15:48:50.340 | INFO     | metagpt.const:get_metagpt_package_root:32 - Package root set to /workspaces/MetaGPT
============================================================ test session starts =============================================================
platform linux -- Python 3.11.4, pytest-7.3.1, pluggy-1.0.0 -- /usr/local/py-utils/venvs/pytest/bin/python
cachedir: .pytest_cache
rootdir: /workspaces/MetaGPT
plugins: mock-3.11.1, anyio-3.7.1, asyncio-0.23.2
asyncio: mode=Mode.STRICT
collected 3 items                                                                                                                            

tests/metagpt/provider/test_zhipuai_api.py::test_zhipuai_completion PASSED
tests/metagpt/provider/test_zhipuai_api.py::test_zhipuai_acompletion PASSED
tests/metagpt/provider/test_zhipuai_api.py::test_zhipuai_proxy PASSED
 ```
**Other**

### DIFF
--- a/metagpt/provider/zhipuai_api.py
+++ b/metagpt/provider/zhipuai_api.py
@@ -50,6 +50,8 @@ class ZhiPuAIGPTAPI(BaseGPTAPI):
         assert config.zhipuai_api_key
         zhipuai.api_key = config.zhipuai_api_key
         openai.api_key = zhipuai.api_key  # due to use openai sdk, set the api_key but it will't be used.
+        if config.openai_proxy:
+            openai.proxy = config.openai_proxy
 
     def _const_kwargs(self, messages: list[dict]) -> dict:
         kwargs = {"model": self.model, "prompt": messages, "temperature": 0.3}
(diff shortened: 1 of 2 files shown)


geekan:
It looks like there's room for optimization, but I'll get through first

springasa:
> It looks like there's room for optimization, but I'll get through first
Thank you for your feedback. I also agree that there is room for optimization, and these changes should be more appropriately placed in the parent class. However, making such a modification would have a significant impact. Therefore, the current changes are addressing the specific issues with the provider first.

springasa:
Thank you for your feedback @geekan. I also agree that there is room for optimization, and these changes should be more appropriately placed in the parent class. However, making such a modification would have a significant impact. Therefore, the current changes are addressing the specific issues with the provider first.

PR #638, opened 2023-12-27
Feat update to pydantic2

better629:
**Features**

- update to pydantic2
- change role's private fields into public
    
**Feature Docs**

**Influence**

- Some pydantic based packages have been checked
- There are no private fields in Role/Action/Environment and so on

**Result**

```
(metagpt_v2) MacBook-Pro:MetaGPT xxx$ pytest -s tests/metagpt/serialize_deserialize/test_*
....
=================================================================== 36 passed, 9 warnings in 148.16s (0:02:28) ===================================================================

```

**Other**

### DIFF
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -26,7 +26,7 @@ action_subclass_registry = {}
 
 
 class Action(BaseModel):
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, exclude=["llm"])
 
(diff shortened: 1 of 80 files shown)


metagpt/actions/action.py
```diff
@@ -19,50 +19,31 @@
     CodeSummarizeContext,
     CodingContext,
     RunCodeContext,
+    SerDeserMixin,
```
geekan, line 22: It is recommended to call it JsonSerializationMixin or SerializationMixin, which is more generic

PR #639, opened 2023-12-27
feat: +unit test

iorisa:
**Features**
+ unit tests


metagpt/actions/prepare_documents.py
```diff
@@ -36,7 +36,8 @@ def _init_repo(self):
         if not path:
             name = CONFIG.project_name or FileRepository.new_filename()
             path = Path(CONFIG.workspace_path) / name
-
+        else:
+            path = Path(CONFIG.project_path)
```
geekan, line 40: This new logic is wrong and contrary to line 35

metagpt/actions/write_prd.py
```diff
@@ -123,7 +124,8 @@ async def _run_new_requirement(self, requirements, schema=CONFIG.prompt_schema)
         #     logger.info(rsp)
         project_name = CONFIG.project_name if CONFIG.project_name else ""
         context = CONTEXT_TEMPLATE.format(requirements=requirements, project_name=project_name)
-        node = await WRITE_PRD_NODE.fill(context=context, llm=self.llm)  # schema=schema
+        write_prd_node = WRITE_PRD_NODE if not project_name else WRITE_PRD_NODE_NO_NAME
```
geekan, line 127: The original intention is to kick out the corresponding key when using, such as fill and exclude

metagpt/tools/ut_writer.py
```diff
@@ -250,21 +255,22 @@ def build_api_doc(self, node: dict, path: str, method: str) -> str:
 
         return doc
 
-    def _store(self, data, base, folder, fname):
+    async def _store(self, data, base, folder, fname):
```
geekan, line 258: Referring to the aread function of the common.py, can I encapsulate an awrite, and is it more convenient to use this function?

metagpt/actions/prepare_documents.py
iorisa: `CONFIG.project_path` comes from `--project-path` argument, str type, '' or a valid path.

metagpt/actions/write_prd.py
iorisa: + ActionNode.fill(..., exclude)

metagpt/actions/prepare_documents.py
geekan: I see what you mean, maybe we should use different variable names? The path that precedes and the path that follows is not of the same type

metagpt/tools/ut_writer.py
iorisa: Replace with common.awrite

PR #651, opened 2023-12-29
Feat fix test_role

better629:
**Features**

- fix test_role
    
**Feature Docs**

**Influence**

**Result**

pytest -s -vv tests/metagpt/test_role.py
================================= 3 passed in 2.43s

**Other**


geekan (approved):
LGTM

PR #652, opened 2023-12-29
Feat add extra role/actions' ser&desr unittest

better629:
**Features**

- add extra role/actions' ser&desr unittest
    
**Feature Docs**

**Influence**

**Result**

**Other**


geekan (approved):
LGTM

Issue #654, opened 2023-12-29
用智普AI运行例子报错


xudengke:
**Bug description**
用智普AI运行例子llm_hello_world.py 报错

**Bug solved method**

**Environment information**

- LLM type and model name:
- System version:windows11
- Python version:3.10

- packages version:
- installation method: 

**Screenshots or logs**
C:/Programs/Python/Python310/python.exe d:/AI/MetaGPT-main/examples/llm_hello_world.py
2023-12-29 16:14:44.661 | INFO     | metagpt.const:get_metagpt_package_root:32 - Package root set to D:\AI\MetaGPT-main 
2023-12-29 16:14:49.173 | INFO     | metagpt.provider.openai_api:update_cost:91 - Total running cost: $0.000 | Max budget: $10.000 | Current cost: $0.000, prompt_tokens: 2, completion_tokens: 60
2023-12-29 16:14:49.174 | INFO     | __main__:main:16 -  Hello! I'm ZhiPuQingYan（智谱清言）, a language model jointly trained by KEG Lab of Tsinghua University and Zhipu AI Company. I'm here to answer your questions and provide help. How can I assist you today?
 Hello! I'm ZhiPuQingYan（智谱清言）, a language model jointly trained by KEG Lab of Tsinghua University and Zhipu AI Company. I'm here to answer your questions and provide help. How can I assist you today?Traceback (most recent call last):
  File "d:\AI\MetaGPT-main\examples\llm_hello_world.py", line 28, in <module>
    asyncio.run(main())
  File "C:\Programs\Python\Python310\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "C:\Programs\Python\Python310\lib\asyncio\base_events.py", line 649, in run_until_complete
    return future.result()
  File "d:\AI\MetaGPT-main\examples\llm_hello_world.py", line 17, in main
    logger.info(await llm.aask_batch(["hi", "write python hello world."]))
  File "C:\Programs\Python\Python310\lib\site-packages\metagpt\provider\base_gpt_api.py", line 76, in aask_batch        
    rsp_text = await self.acompletion_text(context)
  File "C:\Programs\Python\Python310\lib\site-packages\tenacity\_asyncio.py", line 88, in async_wrapped
    return await fn(*args, **kwargs)
  File "C:\Programs\Python\Python310\lib\site-packages\tenacity\_asyncio.py", line 47, in __call__
    do = self.iter(retry_state=retry_state)
[...]

PR #661, opened 2024-01-02
bugfix: fix zhipuapi header str error

cs686:
zhipu api return bytes,but need str. #654


better629:
@cs686 we fix it with https://github.com/geekan/MetaGPT/pull/662 . Thank you~
you can checkout with `dev` branch.

PR #662, opened 2024-01-02
fix issue 654 and re-add system_msg judgement

better629:
**Features**

- fix issue https://github.com/geekan/MetaGPT/issues/654
- re-add system_msg judgement due to deleted by someone else and some llms don't need system_msg.
    
**Feature Docs**

**Influence**

**Result**

**Other**


geekan (approved):
More complete than #661, LGTM

PR #663, opened 2024-01-02
feat: +unit test

iorisa:
**Features**
+ role.py, redis.py, s3.py, schema.py unit test

### DIFF
--- a/metagpt/utils/redis.py
+++ b/metagpt/utils/redis.py
@@ -23,15 +23,7 @@ class Redis:
     async def _connect(self, force=False):
         if self._client and not force:
             return True
-        is_ready = (
-            CONFIG.REDIS_HOST
-            and CONFIG.REDIS_HOST != "YOUR_REDIS_HOST"
-            and CONFIG.REDIS_PORT
-            and CONFIG.REDIS_PORT != "YOUR_REDIS_PORT"
-            and CONFIG.REDIS_DB is not None
-            and CONFIG.REDIS_PASSWORD is not None
-        )
-        if not is_ready:
+        if not self.is_configured:
             return False
 
         try:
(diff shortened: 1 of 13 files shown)


metagpt/utils/redis.py
```diff
@@ -22,7 +23,15 @@ def __init__(self):
     async def _connect(self, force=False):
         if self._client and not force:
             return True
-        if not CONFIG.REDIS_HOST or not CONFIG.REDIS_PORT or CONFIG.REDIS_DB is None or CONFIG.REDIS_PASSWORD is None:
+        is_ready = (
+            CONFIG.REDIS_HOST
+            and CONFIG.REDIS_HOST != "YOUR_REDIS_HOST"
```
geekan, line 28: It is somewhat inconvenient to judge "not configured" in the place where it is used, and if Redis is used in multiple places, it means that the judgment has to be made in multiple places

geekan: It would be more appropriate to have a separate configuration class for parsing, filling in the blanks, and verifying completeness, and making these judgments within each class is a bit trivial

tests/metagpt/utils/test_s3.py
```diff
@@ -41,17 +41,15 @@ async def test_s3():
     res = await conn.cache(data, ".bak", "script")
     assert "http" in res
 
-
-@pytest.mark.asyncio
-async def test_s3_no_error():
+    key = CONFIG.S3_ACCESS_KEY
+    CONFIG.S3_ACCESS_KEY = "YOUR_S3_ACCESS_KEY"
```
geekan, line 45: In a general sense, CONFIG is a read-only class and should not be writable

metagpt/utils/redis.py
iorisa: Relevant parameter checks have been consolidated under `is_configured`.

tests/metagpt/utils/test_s3.py
iorisa: 1. `CONFIG` is a session-scoped context that supports both read and write operations. 
2. `Config` represents a configuration file and is read-only. 
3. The root cause of the issue here is the absence of the concept of a session in the console version. Therefore, the name `CONFIG` is used. 
4. To support the concept of a session functionally, the `contextvars.ContextVar` data type has been employed.

geekan: There's going to be some tricky tricky about that. The associated code needs to be refactored to be more straightforward. If it is a session concept, you should provide a session object, and use session.config or context.config in the session object to accurately express the context in which it is located. It needs to be refactored in the future, so you can put it here first

geekan (approved):
Here are a few legacy issues:
1. Avoid using force push, which here seems to mask some historical commits
2. Reduce the amount of changes to CONFIG, considering that it must be refactored

PR #668, opened 2024-01-03
change unittest result format from junit to plain text

voidking:
**Features**
change unittest result format from junit to plain text


garylin2099:
LGTM

PR #673, opened 2024-01-03
mock serpapi/serper response

shenchucheng:
- mock serpapi/serper response


geekan (changes requested):
I prefer to make a mock that is non-intrusive, i.e. take over the run interface

PR #679, opened 2024-01-04
fixbug: recursive search for provider

iorisa:
**Features**
fixbug: Fix the issue of recursively searching for valid LLM configurations when no LLM is configured.


geekan (approved):
LGTM

PR #685, opened 2024-01-04
Reduce test time with a global LLM mock

garylin2099:
**Features**
- Reduce test time to less than 20mins
- Almost all tests are mocked, meaning possible to automate run on every PR, which will be very useful to discover problems early
- Mock avoid providers folder which allows model APIs to be tested differently

**Result**
https://github.com/garylin2099/MetaGPT/actions/runs/7409829788/job/20160838711

**Limitation**
Still some failed tests (~20), need to solve all of them for the tests to be of truly useful


tests/conftest.py
```diff
@@ -170,6 +151,12 @@ def init_config():
     Config()
 
 
+@pytest.fixture(scope="function")
+def new_filename(mocker):
```
iorisa, line 155: 这个mock放这里，以后上需求分片功能时需要从conftest.py里删掉。
建议代码里加点备注说一下这事。

garylin2099: done

geekan (approved):
LGTM

PR #689, opened 2024-01-05
Fix some bugs

shenchucheng:
- fix test_scrape_web_page proxy error
- fix issue with concatenated prints in llm stream response


geekan (approved):
LGTM

PR #717, opened 2024-01-08
fixbug: role init with is_human=True was not work

kkdev163:
bugfix:  role init with is_human=True was not work

### DIFF
--- a/metagpt/roles/role.py
+++ b/metagpt/roles/role.py
@@ -166,6 +166,9 @@ class Role(SerializationMixin, is_polymorphic_base=True):
         Role.model_rebuild()
         super().__init__(**data)
 
+        if data.get("is_human"):
+            self.llm = HumanProvider()
+
         self.llm.system_prompt = self._get_prefix()
         self._watch(data.get("watch") or [UserRequirement])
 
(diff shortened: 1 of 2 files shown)


kkdev163:
When I was browsing the "Tutorials-Human Engagement" section, I encountered an issue where human intervention was not effective. I hope this issue can be fixed in a timely manner. I am a beginner in Python, so if there is a better way to fix this problem, please advise.  @geekan

metagpt/roles/role.py
```diff
@@ -166,6 +166,9 @@ def __init__(self, **data: Any):
         Role.model_rebuild()
         super().__init__(**data)
 
+        if data.get("is_human"):
```
garylin2099, line 169: Good catch. Can use `if self.is_human` here

kkdev163: Thanks, please review my latest commit

garylin2099:
According test run:

1. Unittest looks fine (you can ignore the errors, the errors are not related to your code)
2. Pre-commit checks fail. There is a formatting issue, potentially the import order in your test_role.py. Check https://github.com/geekan/MetaGPT/blob/main/.pre-commit-config.yaml, run the command `pre-commit run --all-files` and commit your format change

kkdev163:
> pre-commit run --all-files and commit your format change

Thanks for giving guide. Please help me review once more.

garylin2099:
LGTM

PR #718, opened 2024-01-08
refactor: test_redis.py

iorisa:
**Features**
- refactor unit test code


geekan (approved):
Notice that unittest.mock is used here, is it optimal?

PR #734, opened 2024-01-10
add action_outcls decorator to support init same class with same class name and fields

better629:
**Features**

- add action_outcls decorator to support init same class with same class name and fields
- add its unittest and update other ut
    
**Feature Docs**

**Influence**

Warning, class from `create_model` with same class name and field definition will be same when comparing.

**Result**

```
(metagpt_v2) MacBook-Pro:MetaGPT xxx$ pytest -s -vv tests/metagpt/serialize_deserialize/test_*

=================================================================== 0 failed, 43 passed, 7 warnings in 57.60s ===================================================================
```

**Other**


geekan (approved):
LGTM

PR #756, opened 2024-01-15
add:openai text to speech

Justin-ZL:
**Features**

- add openai's text to speech in OpenAILLM
    
**Feature Docs**

**Influence**

**Result**

**Other**


geekan (approved):
LGTM

PR #768, opened 2024-01-17
feat: Merge the modifications resulting from integrating with the agent store.

iorisa:
**Features**
- fixbug: `Role.think()`
- rename `first_action` to `action_description`
- update `numpy==1.24.3` to `numpy>=1.24.3`
- update version to "0.7.0"

### DIFF
--- a/metagpt/document_store/faiss_store.py
+++ b/metagpt/document_store/faiss_store.py
@@ -37,11 +37,7 @@ class FaissStore(LocalStore):
         return FAISS.load_local(self.raw_data_path.parent, self.embedding, self.fname)
 
     def _write(self, docs, metadatas):
-        try:
-            store = FAISS.from_texts(docs, self.embedding, metadatas=metadatas)
-        except Exception as e:
-            logger.error(f"Failed to write. error: {e}")
-            raise e
+        store = FAISS.from_texts(docs, self.embedding, metadatas=metadatas)
         return store
 
     def persist(self):
--- a/metagpt/document_store/faiss_store.py
+++ b/metagpt/document_store/faiss_store.py
@@ -37,7 +37,11 @@ class FaissStore(LocalStore):
         return FAISS.load_local(self.raw_data_path.parent, self.embedding, self.fname)
 
     def _write(self, docs, metadatas):
-        store = FAISS.from_texts(docs, self.embedding, metadatas=metadatas)
+        try:
+            store = FAISS.from_texts(docs, self.embedding, metadatas=metadatas)
+        except Exception as e:
+            logger.error(f"Failed to write. error: {e}")
+            raise e
         return store
 
     def persist(self):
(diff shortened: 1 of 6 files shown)


metagpt/document_store/faiss_store.py
```diff
@@ -37,7 +37,11 @@ def _load(self) -> Optional["FaissStore"]:
         return FAISS.load_local(self.raw_data_path.parent, self.embedding, self.fname)
 
     def _write(self, docs, metadatas):
-        store = FAISS.from_texts(docs, self.embedding, metadatas=metadatas)
+        try:
```
geekan, line 40: Use handle_exception decorator if needed. But I still want to ask, is it really necessary to catch anomalies in scenes like this?

metagpt/roles/role.py
```diff
@@ -524,6 +524,7 @@ def is_idle(self) -> bool:
 
     async def think(self) -> Action:
         """The exported `think` function"""
+        await self._observe()
```
geekan, line 527: why _observe here?

metagpt/roles/role.py
```diff
@@ -533,9 +534,14 @@ async def act(self) -> ActionOutput:
         return ActionOutput(content=msg.content, instruct_content=msg.instruct_content)
 
     @property
-    def first_action(self) -> str:
-        """AgentStore uses this attribute to display to the user what actions the current role should take."""
-        # FIXME: this is a hack, we should not use the first action to represent the todo
+    def action_description(self) -> str:
+        """
+        AgentStore uses this attribute to display to the user what actions the current role should take.
```
geekan, line 539: This naming loses the meaning of the first action, the handling is implicit and needs to be commented in the function

metagpt/roles/role.py
geekan: pending test

geekan: There are still some counter-instincts in the think _observe, can you add the reason to the comments?

metagpt/roles/role.py
geekan: The problem here is that action_description doesn't express what strategy it wants to use to select actions, and the code lacks corresponding comments. In particular, it seems that there should be a unified "function that is about to execute an action" before converting it to str, otherwise developers will be confused about which action to execute and where

iorisa: `Role` provides the default property, and this property can be overridden by children classes.
Take `Engineer` as example:
```
class Engineer(Role):
    @property
    def action_description(self) -> str:
        """AgentStore uses this attribute to display to the user what actions the current role should take."""
        return self.next_todo_action
        
async def _act(self) -> Message | None:
        """Determines the mode of action based on whether code review is used."""
        if self.rc.todo is None:
            return None
        if isinstance(self.rc.todo, WriteCode):
            self.next_todo_action = any_to_name(SummarizeCode)
            return await self._act_write_code()
        if isinstance(self.rc.todo, SummarizeCode):
            self.next_todo_action = any_to_name(WriteCode)
            return await self._act_summarize()
        return None
```

geekan (approved):

PR #775, opened 2024-01-18
Fix test generation

arnaudgelas:
* Stop generating unit tests for non-python code files (e.g. ReadMe.md)
* Stop trying to install dependencies when there are none

### DIFF
--- a/metagpt/actions/run_code.py
+++ b/metagpt/actions/run_code.py
@@ -16,6 +16,7 @@
             class.
 """
 import subprocess
+from pathlib import Path
 from typing import Tuple
 
 from pydantic import Field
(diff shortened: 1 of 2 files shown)


metagpt/actions/run_code.py
```diff
@@ -152,11 +153,23 @@ def _install_via_subprocess(cmd, check, cwd, env):
         return subprocess.run(cmd, check=check, cwd=cwd, env=env)
 
     @staticmethod
-    def _install_dependencies(working_directory, env):
+    def _install_requirements(self, working_directory, env):
```
geekan, line 156: @staticmethod will ignore this self. is it a bug?

metagpt/actions/run_code.py
```diff
@@ -152,11 +153,23 @@ def _install_via_subprocess(cmd, check, cwd, env):
...
+        file_path = os.path.join(working_directory, "requirements.txt")
```
geekan, line 157: use pathlib.Path

metagpt/actions/run_code.py
arnaudgelas: my bad, thanks for pointing out :+1:

metagpt/actions/run_code.py
arnaudgelas: sure

geekan (approved):
LGTM

geekan:
While the improvement in test coverage is lacking here, it does provide meaningful logic

geekan:
@arnaudgelas If the PR is marked as draft, then it cannot be merged

apofic:
Won't this impact non-python unit testing, such as Javascript/Typescript?

geekan (approved):

geekan:
@apofic Yes. This means that there will be missing tests for other languages. Perhaps we need a complete solution, but currently I'm not sure if it exists.

PR #776, opened 2024-01-19
bugfix: DEFAULT_PROVIDER config is invalid

voidking:
1. As long as `self._is_valid_llm_key(self.OPENAI_API_KEY) and not self.OPENAI_API_TYPE and self.OPENAI_API_MODEL` is true, LLM will use openai, `DEFAULT_PROVIDER` is invalid.
2. If users want to change to other LLM, they have to comment out `OPENAI_API_KEY` in the config if they have set it before.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -4,6 +4,9 @@
 #### Project Path Setting
 # WORKSPACE_PATH: "Path for placing output files"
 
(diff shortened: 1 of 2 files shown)


voidking:
The dev branch can avoid the problem. So I will close it.

PR #784, opened 2024-01-22
feat: Replace `CONTEXT` with `Context()`

iorisa:
**Features**
feat: Replace `CONTEXT` with `Context()`
  - The default `CONTEXT` is equal to `Context()`.
  - Due to the unrestricted read and write access to the global variable `CONTEXT`, it resulted in uncontrolled propagation of changes. Therefore, this global variable has been deprecated.

From now, the `Context` object is uniformly created during scene initialization and implicitly passed through the object hierarchy. Examples can be found in the `generate_repo` function in `startup.py`.


geekan (approved):
LGTM

PR #797, opened 2024-01-25
fixbug: compatible with windows path

iorisa:
**Features**
- Windows use `\` as path delimiter, while Unix use `/` as path delimiter. This causes issues with `dependencies.json` not working consistently across different platforms.
- Standardize the path delimiter for `.dependencies.json` to `/`
- Add the `filter_ignored` parameter to `FileRepository.get_all()` to enable the functioning of `.gitignore` filters.


geekan (approved):
LGTM

PR #804, opened 2024-01-29
Enhance local codebase reader function

maxenergy:
**Features**

- xx
- yy
    
**Feature Docs**

**Influence**

**Result**

**Other**


geekan:
What is it designed to do? What does it actually do?

PR #825, opened 2024-02-02
fixbug: compatible with windows path

mannaandpoem:
**Features**
fixbug: compatible with windows path

### DIFF
--- a/metagpt/utils/dependency_file.py
+++ b/metagpt/utils/dependency_file.py
@@ -96,7 +96,8 @@ class DependencyFile:
             key = Path(filename).relative_to(root)
         except ValueError:
             key = filename
-        return set(self._dependencies.get(str(key), {}))
+        key = re.sub(r"\\+", "/", str(key))
+        return set(self._dependencies.get(key, {}))
 
     def delete_file(self):
         """Delete the dependency file."""
--- a/tests/metagpt/test_incremental_dev.py
+++ b/tests/metagpt/test_incremental_dev.py
@@ -45,6 +45,7 @@ PROJECT_NAMES = [
 ]
 
 
+@pytest.mark.skip
 def test_simple_add_calculator():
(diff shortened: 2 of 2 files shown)


metagpt/utils/dependency_file.py
```diff
@@ -96,7 +96,8 @@ async def get(self, filename: Path | str, persist=True):
             key = Path(filename).relative_to(root)
         except ValueError:
             key = filename
-        return set(self._dependencies.get(str(key), {}))
+        key = re.sub(r"\\+", "/", str(key))
```
geekan, line 99: 1. use `pathlib.path(r'foo\bar').as_posix()` would be better.
2. it's too many defense in this file. I see 4 `re.sub(r"\\+"`, it's not necessary. Just delete 3 and leave one path.as_posix()

tests/metagpt/test_incremental_dev.py
```diff
@@ -115,8 +116,11 @@ def get_incremental_dev_result(idea, project_name, use_review=True):
     if not project_path.exists():
         # If the project does not exist, extract the project file
         try:
-            # Use the tar command to extract the .zip file
-            subprocess.run(["tar", "-xf", f"{project_path}.zip", "-C", str(project_path.parent)], check=True)
+            if os.name == "nt":
+                subprocess.run(["tar", "-xf", f"{project_path}.zip", "-C", str(project_path.parent)], check=True)
+            else:
+                subprocess.run(["unzip", f"{project_path}.zip", "-d", str(project_path.parent)], check=True)
```
geekan, line 122: I think a generic unzip function, or something like that, should be provided. This function should check if the command exists and then execute the command, a process that is not system-agnostic, but the command provided

geekan (changes requested):

geekan (approved):
lgtm

PR #833, opened 2024-02-02
Fix `Role` not fully defined error debugger

shenchucheng:
**Features**
- import Environment/Role in TYPE_CHECKING mode
- fix `Role` not fully defined error accounting to https://stackoverflow.com/questions/63420889/fastapi-pydantic-circular-references-in-separate-files


iorisa:
LGTM

geekan (approved):
lgtm

PR #851, opened 2024-02-06
feat: add parse arguments for openai tool_calls.

orange-crow:
**Features**

add parse arguments for openai tool_calls.

### DIFF
--- a/metagpt/provider/openai_api.py
+++ b/metagpt/provider/openai_api.py
@@ -8,6 +8,7 @@
 """
 
 import json
+import re
 from typing import AsyncIterator, Optional, Union
 
 from openai import APIConnectionError, AsyncOpenAI, AsyncStream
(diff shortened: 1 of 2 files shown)


metagpt/provider/openai_api.py
```diff
@@ -209,7 +234,14 @@ def get_choice_function_arguments(self, rsp: ChatCompletion) -> dict:
...
             # reponse is code
-            return json.loads(message.tool_calls[0].function.arguments, strict=False)
+            try:
+                return json.loads(message.tool_calls[0].function.arguments, strict=False)
+            except json.decoder.JSONDecodeError as e:
+                error_msg = (
+                    f"Got JSONDecodeError for \n{'--'*40} \n{message.tool_calls[0].function.arguments}, {str(e)}"
+                )
+                logger.error(error_msg)
+                return self._parse_arguments(message.tool_calls[0].function.arguments)
```
geekan, line 244: I don't think that's a good approach. We should either reuse the repair-related method, or we should retry until it succeeds

geekan (changes requested):

metagpt/provider/openai_api.py
orange-crow: Retry until it succeeds sounds good.

orange-crow: I try it like this, but it does not work.
image
image

And,  I found these in openai functions:
image
When parsing the results of function calls, it seems necessary to use regular expressions for matching.

garylin2099:
severely impacts the ability of using openai response, merged first, recorded todo

PR #855, opened 2024-02-06
fix critical CodeInterpreter error: empty code when aask_code not returning language

garylin2099:
An important bug fix to CodeInterpreter and needs to be merged into main directly, independent of other updates in CodeInterpreter branch.

Previously, when markdown code is generated, we manually set the code to be empty and maintain the rendered result only. This sometimes set real python code to be empty by mistake, if LLM function call rsp does not return a language value. In addition, the logic of empty string setting is not easily understood. Hence, we removed it


garylin2099:
now ready yet

geekan (approved):
lgtm

PR #922, opened 2024-02-21
Update role.py : Role.set_actions()

lixiangtiandashen:
由于提示词里要求llm选择的状态码是 0 到 状态数-1，所以我想这里往states里添加动作对应状态时，也应当从0开始。
Due to the requirement in the prompt word that llm selects a status code from 0 to the number of states -1, I believe that when adding action corresponding states to states, it should also start from 0.


geekan (commented):
lgtm

PR #950, opened 2024-02-29
doc: add docker installation section and fold it

voidking:
**Feature Docs**
add docker installation section and fold it, refer to https://github.com/evalplus/evalplus
image


geekan (commented):
LGTM

Issue #733, opened 2024-01-10
LLM may generate comments inside [CONTENT][/CONTENT]  , which causes parsing the JSON to fail.


kkdev163:
**Bug description**
```
parse json from content inside [CONTENT][/CONTENT] failed at retry 1, exp: Expecting ',' delimiter: line 6 column 27 (char 135)
```

**Bug solved method**

Perhaps we could consider adding a constraint to the prompt, indicating not to generate comments inside [CONTENT][/CONTENT], or alternatively, we could trim the comments from the LLM's output.

**Environment information**

- LLM type and model name:  OPENAI gpt-4-1106-preview
- System version: macos 12.5.1
- Python version: python 3.9

- packages version: metagpt commit 82a5eec72707dee44174eae8f8ff1490a6819ecd
- installation method:  pip install from source

**Screenshots or logs**

```
[CONTENT]
{
    "Required Python packages": [
        "numpy==1.21.2",
        "Kivy==2.0.0",
        "pygame==2.0.1",
        "sqlite3==2.6.0"  # sqlite3 is included in Python's standard library, but versioning is for consistency
    ],
    "Required Other language third-party packages": [
        "No third-party dependencies required"
    ],
    "Logic Analysis": [
        [
            "game.py",
            "Contains Game class with core game logic, uses numpy for array manipulation, and interacts with UI and Storage classes"
        ],
        [
            "main.py",
            "Contains main function, initializes the game by calling start_new_game() from Game class"
        ],
        [
            "ui.py",
            "Contains UI class for user interface, uses Kivy for rendering, and interacts with Game class"
        ],
        [
            "storage.py",
            "Contains Storage class for saving and loading high scores using SQLite"
        ]
    ],
    "Task list": [
        "storage.py",
        "game.py",
        "ui.py",
        "main.py"
    ],
    "Full API spec": "",
    "Shared Knowledge": "'game.py' contains the Game class which is central to the game logic and is used by both 'ui.py' for rendering the game state and 'storage.py' for saving the high score.",
    "Anything UNCLEAR": "The monetization strategy for the game is not specified. Will the game include ads, in-app purchases, or be a paid app? This will affect the design of the user interface and potentially the choice of libraries or frameworks."
}
[/CONTENT]
[...]

PR #963, opened 2024-03-05
feat: fix JSONDecodeError error caused by incorrect indentation used …

moyitech:
**Features**

Issue: https://github.com/geekan/MetaGPT/issues/733
c57e4e482e5c370d78e525e50867bc6

We often encounter this issue when using ZHIPU AI instead of OpenAI
Although the repair_llm.output option is enabled in the configuration file, the JSON correction function cannot be used.

**Feature Docs**
By reading config2. py, I learned that the repair_llm output option in config2. yaml should be placed in the first level indentation instead of under llm, so I modified config2. example. yaml

**Result**
db78db8aec4898731bfd31f715ca1a2

### DIFF
--- a/config/config2.example.yaml
+++ b/config/config2.example.yaml
@@ -3,7 +3,6 @@ llm:
   base_url: "YOUR_BASE_URL"
   api_key: "YOUR_API_KEY"
   model: "gpt-4-turbo-preview"  # or gpt-3.5-turbo-1106 / gpt-4-1106-preview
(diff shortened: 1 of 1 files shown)


iorisa:
Good job.

```python
class Config(CLIParams, YamlModel):
    """Configurations for MetaGPT"""

    # Key Parameters
    llm: LLMConfig

    ...

    # Misc Parameters
    repair_llm_output: bool = False
```

PR #983, opened 2024-03-11
feat: merge v0.7.6 to main

iorisa:
**Features**
merge v0.7.6 to main


examples/di/ocr_receipt.py
```diff
@@ -6,9 +6,7 @@ async def main():
     image_path = "image.jpg"
     language = "English"
     requirement = f"""This is a {language} receipt image.
-    Your goal is to perform OCR on images using PaddleOCR, output text content from the OCR results and discard 
-    coordinates and confidence levels, then recognize the total amount from ocr text content, and finally save as table. 
-    Image path: {image_path}.
+    Your goal is to perform OCR on images using PaddleOCR, then extract the total amount from ocr text results, and finally save as table. Image path: {image_path}.
```
geekan, line 9: Is this expected?

metagpt/provider/google_gemini_api.py
```diff
@@ -51,6 +52,10 @@ def __init__(self, config: LLMConfig):
         self.llm = GeminiGenerativeModel(model_name=self.model)
 
     def __init_gemini(self, config: LLMConfig):
+        if config.proxy:
+            logger.info(f"Use proxy: {config.proxy}")
+            os.environ["HTTP_PROXY"] = config.proxy
+            os.environ["HTTP_PROXYS"] = config.proxy
```
geekan, line 58: Is this expected?

iorisa: `GenerativeModel` does not provide an interface to set up the http proxy:
```python
def configure(
    *,
    api_key: str | None = None,
    credentials: ga_credentials.Credentials | dict | None = None,
    # The user can pass a string to choose `rest` or `grpc` or 'grpc_asyncio'.
    # See `_transport_registry` in `DiscussServiceClientMeta`.
    # Since the transport classes align with the client classes it wouldn't make
    # sense to accept a `Transport` object here even though the client classes can.
    # We could accept a dict since all the `Transport` classes take the same args,
    # but that seems rare. Users that need it can just switch to the low level API.
    transport: str | None = None,
    client_options: client_options_lib.ClientOptions | dict | None = None,
    client_info: gapic_v1.client_info.ClientInfo | None = None,
    default_metadata: Sequence[tuple[str, str]] = (),
):
```
```python
class ClientOptions(object):
    """Client Options used to set options on clients.

    Args:
        api_endpoint (Optional[str]): The desired API endpoint, e.g.,
            compute.googleapis.com
        client_cert_source (Optional[Callable[[], (bytes, bytes)]]): A callback
            which returns client certificate bytes and private key bytes both in
            PEM format. ``client_cert_source`` and ``client_encrypted_cert_source``
            are mutually exclusive.
        client_encrypted_cert_source (Optional[Callable[[], (str, str, bytes)]]):
            A callback which returns client certificate file path, encrypted
            private key file path, and the passphrase bytes.``client_cert_source``
            and ``client_encrypted_cert_source`` are mutually exclusive.
        quota_project_id (Optional[str]): A project name that a client's
            quota belongs to.
        credentials_file (Optional[str]): A path to a file storing credentials.
            ``credentials_file` and ``api_key`` are mutually exclusive.
        scopes (Optional[Sequence[str]]): OAuth access token override scopes.
        api_key (Optional[str]): Google API key. ``credentials_file`` and
            ``api_key`` are mutually exclusive.
        api_audience (Optional[str]): The intended audience for the API calls
            to the service that will be set when using certain 3rd party
            authentication flows. Audience is typically a resource identifier.
            If not set, the service endpoint value will be used as a default.
            An example of a valid ``api_audience`` is: "https://language.googleapis.com".
        universe_domain (Optional[str]): The desired universe domain. This must match
            the one in credentials. If not set, the default universe domain is
            `googleapis.com`. If both `api_endpoint` and `universe_domain` are set,
            then `api_endpoint` is used as the service endpoint. If `api_endpoint` is
            not specified, the format will be `{service}.{universe_domain}`.

    Raises:
        ValueError: If both ``client_cert_source`` and ``client_encrypted_cert_source``
            are provided, or both ``credentials_file`` and ``api_key`` are provided.
    """

    def __init__(
        self,
        api_endpoint=None,
        client_cert_source=None,
        client_encrypted_cert_source=None,
        quota_project_id=None,
        credentials_file=None,
        scopes=None,
        api_key=None,
        api_audience=None,
        universe_domain=None,
    ):
    ```

examples/di/ocr_receipt.py
iorisa: fixed. reset to `main` branch

geekan (approved):

PR #985, opened 2024-03-11
feat: + tree command

iorisa:
**Features**
Implement the same functionality as the `tree` command.
Example:
```
        root
        +-- dir1
        |   +-- file1.txt
        |   +-- file2.txt
        +-- dir2
        |   +-- subdir1
        |   |   +-- file1.txt
        |   |   +-- file2.txt
        |   +-- subdir2
        |       +-- file1.txt
        |       +-- file2.txt
        +-- file.txt
```

### DIFF
--- a/metagpt/utils/tree.py
+++ b/metagpt/utils/tree.py
@@ -6,94 +6,166 @@
 @File    : tree.py
 @Desc    : Implement the same functionality as the `tree` command.
 Example:
-        root
-        +-- dir1
-        |   +-- file1.txt
-        |   +-- file2.txt
-        +-- dir2
-        |   +-- subdir1
-        |   |   +-- file1.txt
-        |   |   +-- file2.txt
-        |   +-- subdir2
-        |       +-- file1.txt
-        |       +-- file2.txt
-        +-- file.txt
+        Usage:
+            >>> print_tree(".")
+            utils
+            +-- serialize.py
+            +-- project_repo.py
+            +-- tree.py
+            +-- mmdc_playwright.py
+            +-- dependency_file.py
+            +-- index.html
+            +-- make_sk_kernel.py
+            +-- token_counter.py
+            +-- embedding.py
+            +-- repair_llm_raw_output.py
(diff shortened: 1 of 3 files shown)


metagpt/utils/tree.py
```diff
@@ -0,0 +1,99 @@
...
+        |       +-- file1.txt
+        |       +-- file2.txt
+        +-- file.txt
+"""
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Callable, Dict, List
+
+from anthropic import BaseModel
```
geekan, line 27: ?

metagpt/utils/tree.py
```diff
@@ -0,0 +1,99 @@
...
+class Tree(BaseModel):
+    """
+    Represents a directory tree structure.
+
+    Attributes:
+        root (str): The root directory of the tree.
+        tree (Dict[str, Dict]): The tree structure as a dictionary.
+
+    Methods:
+        print: Print the directory tree structure.
```
geekan, line 40: give examples

metagpt/utils/tree.py
```diff
@@ -0,0 +1,99 @@
...
+
+    """
+
+    root: str
+    tree: Dict[str, Dict] = Field(default_factory=dict)
```
geekan, line 45: Set as private variable

geekan:
I still want to ask a question: now that the tree command exists and it provides a large number of options, why should we reimplement it? What are the benefits?

metagpt/utils/tree.py
iorisa: Fixed.
Classes have been simplified into functions.

metagpt/utils/tree.py
iorisa: done

metagpt/utils/tree.py
iorisa: Fixed.
Classes have been simplified into functions

geekan (approved):

Issue #979, opened 2024-03-11
`utils.text.reduce_message_length` Not reducing text length


azurewtl:
**Bug description**
I come across this issue.
```python
  File "/Users/azure/Documents/Workspace/Datasci/lib/python3.10/site-packages/metagpt/utils/text.py", line 31, in reduce_message_length
    raise RuntimeError("fail to reduce message length")
RuntimeError: fail to reduce message length
```

**Bug solved method**
Digging into the code, I assume `utils.text.reduce_message_length()` only check if the token is short enough.
If it's too long, it simply raise exception, instead of shorten it.
Following it the code in `utils.text.reduce_message_length()`
```python
def reduce_message_length(
    msgs: Generator[str, None, None],
    model_name: str,
    system_text: str,
    reserved: int = 0,
) -> str:
    max_token = TOKEN_MAX.get(model_name, 2048) - count_string_tokens(system_text, model_name) - reserved
    for msg in msgs:
        if count_string_tokens(msg, model_name) < max_token or model_name not in TOKEN_MAX:
            return msg

    raise RuntimeError("fail to reduce message length")
``` 

- LLM type and model name:
- System version:MetaGPT 0.7.4
- Python version: Python 3.10.13

Is it a feature that is not implemented yet, or I can try to create a PR to fix it

iorisa:
Thank you for your feedback.
The issue isn't caused by `reduce_message_length`, but rather by incorrect usage of the `reserved` parameter when using `reduce_message_length`.
In the v0.6-release branch, the source code looks like this：
```python
prompt = reduce_message_length(gen_msg(), model_name, system_text, CONFIG.max_tokens_rsp)
```
However, in the v0.7-release branch, the source code looks like this：
```python
prompt = reduce_message_length(gen_msg(), model_name, system_text, 4096)
```
The `reserved` parameter is assigned a constant value of 4096, which could result in `max_token = TOKEN_MAX.get(model_name, 2048) - count_string_tokens(system_text, model_name) - reserved` being constantly less than 0.
You can submit a PR to replace `4096` with `config.llm.max_token`：
<img width="937" alt="截屏2024-03-11 15 27 32" src="https://github.com/geekan/MetaGPT/assets/46912756/f1769d9c-07ff-4127-a88b-58925716356b">
[...]

PR #986, opened 2024-03-11
fixbug#979 : `RuntimeError: fail to reduce message length` 

azurewtl:
#979 
**Features**
- updated MAX-TOKEN according to openai document
- minior optimization of code style
- fixed issues `RuntimeError: fail to reduce message length`
   
**Influence**
before gpt-3.5-turbo will raise error, this pr fixed it


geekan (approved):
lgtm

PR #995, opened 2024-03-12
Restructure DataInterpreter

garylin2099:
1. Merged the Interpreter and MLE into one, unified name as Data Interpreter.
2. Merged multiple write code actions into one:
   - Refactored WriteAnalysisCode to make it easier to understand. After refactoring, it's clear that WriteAnalysisCode requires four key pieces of information, each independent and originating from a specific module. WriteAnalysisCode merely combines these pieces of information for code generation:
      1. User Requirement, mandatory, obtained from Role.
      2. Plan Status, optional, obtained from Planner if available.
      3. Tool Info, optional, obtained from ToolRecommender if available.
      4. Working Memory, optional, encompasses trial and error, review, etc., obtained from Role.
3. Decoupled the plan and tool module. Added a tool recommender. Simplified and optimized tool use:
   - The original code logic involved the planner breaking down tasks and selecting tools based on task_type, which made tool usage unnecessarily dependent on the planner. The planner and tool should be independent modules.
   - Upon closer examination, the original tool type usage prompt was actually a task type prompt, incorporating human priors for specific task types rather than tool use. Hence, the tool type was discarded in favor of defining task types.
   - Changed the tool side to use text tags. Considering a tool can be used for different types and granularities of tasks, it allows for a set of tags to facilitate user specification or retrieval. Example usage:
      ```python
      @register_tool(tags=["data preprocessing", "machine learning"])
      class FillMissingValue(DataPreprocessTool):
          ...
      
      @register_tool(tags=["feature engineering", "machine learning"])
      class PolynomialExpansion(MLProcess):
          ...
      
      # The above tools can be specified in the following ways:
      
      di = DataInterpreter(tools=["FillMissingValue"]) # Specify by name.
      di = DataInterpreter(tools=["machine learning"]) # Specify by any tag.
      di = DataInterpreter(tools=["<all>"]) # Specify the full set of tools, relying on ToolRecommender for recommendations.
      di = DataInterpreter(tools=["data preprocessing", "PolynomialExpansion"]) # Mix of names and tags, meaning the set of available tools includes all preprocessing tools plus one PolynomialExpansion feature engineering tool.

      di = DataInterpreter() # By default, no tools are used.
      ```
   - Introduced the ToolRecommender abstraction to take over tool recommendation, capable of outputting a Tool Info string directly usable for downstream prompts. The default process involves recall + rank, with default rank using LLM rank and recall being extendable to different subclasses, such as TypeMatch, BM25, etc.
   - Simplified tool schema generation. Initially, it was in the format of an OpenAI function dictionary, but as function calls are not made, this format was not strictly necessary. Moreover, since the original tool schema generation heavily depended on docstrings, which were not robust and difficult to maintain due to a series of regular expressions, this approach was discarded. Now, function signatures are used, with docstrings as supplementary, testing shows no issues with effectiveness. For the format of new tool schema, check `test_tool_convert.py`
4. Changed code generation methods to support LLMs other than OpenAI.
   - Previously used aask_code, which is based on function calls and tested to be less stable than aask, and other LLMs may not necessarily support function calls. Hence, shifted to using aask.
5. Supported ReAct mode in Interpreter.
   - ReAct mode has advantages in some simple scenarios or when more flexible trial and error is needed, thus it has been introduced to the Interpreter.


examples/di/machine_learning_complex.py
```diff
@@ -1,10 +1,10 @@
 import asyncio
 
-from metagpt.roles.di.ml_engineer import MLEngineer
+from metagpt.roles.di.data_interpreter import DataInterpreter
 
 
 async def main(requirement: str):
-    role = MLEngineer(auto_run=True, use_tools=True)
+    role = DataInterpreter(use_reflection=True, tools=["<all>"])
```
geekan, line 7: This is a very specific usage convention, and it should be very clearly stated how to use it.

metagpt/actions/di/write_analysis_code.py
```diff
@@ -6,150 +6,70 @@
...
         )
-        context.append(Message(content=tools_instruction, role="user"))
 
-        # prepare prompt & LLM call
-        prompt = self.insert_system_message(context)
-        tool_config = create_func_call_config(CODE_GENERATOR_WITH_TOOLS)
-        rsp = await self.llm.aask_code(prompt, **tool_config)
+        working_memory = working_memory or []
+        context = [Message(content=structual_prompt, role="user")] + working_memory
+        context = process_message(context)
```
geekan, line 54: I prefer not to split into two lines

metagpt/actions/di/write_analysis_code.py
```diff
@@ -6,150 +6,70 @@
...
+
+        # LLM call
+        if not use_reflection:
```
geekan, line 57: I tend to use if use_reflection

metagpt/actions/di/write_analysis_code.py
```diff
@@ -6,150 +6,70 @@
...
+            rsp = await self.llm.aask(context, system_msgs=[INTERPRETER_SYSTEM_MSG], **kwargs)
+            code = CodeParser.parse_code(block=None, text=rsp)
+
```
geekan, line 60: Remove redundant blank lines

metagpt/roles/di/data_interpreter.py
```diff
@@ -72,14 +136,49 @@ async def _write_and_exec_code(self, max_retry: int = 3):
...
+                TaskType.FEATURE_ENGINEERING.type_name,
+                TaskType.MODEL_TRAIN.type_name,
+            ]
+        ):
+            return
+        logger.info("Check updated data")
+        code = await CheckData().run(self.planner.plan)
+        if not code.strip():
+            return
+        success = False
```
geekan, line 179: It's not necessary

metagpt/roles/di/data_interpreter.py
```diff
@@ -72,14 +136,49 @@ async def _write_and_exec_code(self, max_retry: int = 3):
...
+            or self.planner.plan.current_task.task_type
+            not in [
+                TaskType.DATA_PREPROCESS.type_name,
+                TaskType.FEATURE_ENGINEERING.type_name,
+                TaskType.MODEL_TRAIN.type_name,
+            ]
+        ):
+            return
+        logger.info("Check updated data")
+        code = await CheckData().run(self.planner.plan)
```
geekan, line 176: Pass in the necessary parameters when instantiating the action to complete the correct priority setting of the context / config

metagpt/roles/di/data_interpreter.py
```diff
@@ -72,14 +136,49 @@ async def _write_and_exec_code(self, max_retry: int = 3):
...
 
-    async def _write_code(self):
-        todo = WriteCodeWithoutTools() if not self.use_tools else WriteCodeWithTools(selected_tools=self.tools)
+    async def _write_code(
+        self,
+        counter,
+        plan_status="",
+        tool_info="",
+    ):
+        todo = WriteAnalysisCode()
```
geekan, line 147: Pass in the necessary parameters when instantiating the action to complete the correct priority setting of the context / config

metagpt/roles/di/data_interpreter.py
```diff
@@ -51,14 +99,30 @@ async def _write_and_exec_code(self, max_retry: int = 3):
...
+            context = (
+                self.working_memory.get()[-1].content if self.working_memory.get() else ""
+            )  # thoughts from _think stage in 'react' mode
+            plan = self.planner.plan if self.use_plan else None
+            tool_info = await self.tool_recommender.get_recommended_tool_info(context=context, plan=plan)
+        else:
+            tool_info = ""
+
+        # data info
+        await self._check_data()
```
geekan, line 116: I still recommend writing similar functions in a reentrant form rather than with state.

metagpt/tools/tool_recommend.py
```diff
@@ -0,0 +1,216 @@
...
+- Output a json list of tool names:
+```json
+["tool_name1", "tool_name2", ...]
+```
+"""
+
+
+class RecommendTool(Action):
+    async def run(self, prompt):
+        return await self._aask(prompt)
```
geekan, line 54: Unnecessary abstraction, recommended to remove

metagpt/tools/tool_recommend.py
```diff
@@ -0,0 +1,216 @@
...
+    2. Rank: Use LLM to select final candidates from recalled set.
+    """
+
+    tools: dict[str, Tool] = {}
+    force: bool = False  # whether to forcedly recommend the specified tools
+
+    @field_validator("tools", mode="before")
+    @classmethod
+    def validate_tools(cls, v: list[str]) -> dict[str, Tool]:
+        if v == ["<all>"]:
```
geekan, line 70: This is an convention. I don't tend to use conventions to solve problems

metagpt/roles/di/data_interpreter.py
garylin2099: Since _check_data is not meant to be used independently or externally, but only encapsulated to hide details and make this step clearer in main pipeline, would passing args be a bit unnecessary?

metagpt/roles/di/data_interpreter.py
garylin2099: Corrected WriteAnalysisCode. For this action, since we want rc.todo to be continuously WriteAnalysisCode, it is a bit inconvenient to set todo to CheckData and back to WriteAnalysisCode

metagpt/tools/tool_recommend.py
garylin2099: Comments added. Will also introduce it on doc sites

garylin2099:
Test coverage 83%, no less than main. No failures other than those existed on main.
PR: https://github.com/geekan/MetaGPT/actions/runs/8267579269/job/22618337523?pr=995
Main: https://github.com/geekan/MetaGPT/actions/runs/8265922605/job/22612737642

geekan (approved):

metagpt/tools/libs/data_preprocess.py
```diff
@@ -16,9 +17,8 @@
 )
 
 from metagpt.tools.tool_registry import register_tool
-from metagpt.tools.tool_type import ToolType
 
-TOOL_TYPE = ToolType.DATA_PREPROCESS.type_name
+TAGS = ["data preprocessing", "machine learning"]
```
better629, line 21: seems better to use enum. so can be used in other files under `libs`

metagpt/tools/tool_recommend.py
better629: how about use all tools if tools is {}

metagpt/tools/tool_registry.py
```diff
@@ -10,52 +10,36 @@
...
 class ToolRegistry(BaseModel):
     tools: dict = {}
-    tool_types: dict = {}
-    tools_by_types: dict = defaultdict(dict)  # two-layer k-v, {tool_type: {tool_name: {...}, ...}, ...}
-
-    @field_validator("tool_types", mode="before")
-    @classmethod
-    def init_tool_types(cls, tool_types: ToolType):
-        return {tool_type.type_name: tool_type.value for tool_type in tool_types}
+    tools_by_tags: dict = defaultdict(dict)  # two-layer k-v, {tag: {tool_name: {...}, ...}, ...}
```
better629, line 26: seems it no need to add additional field `tools_by_tags`. it can be constructed by `self.tools` when using due to the construction speed is usually quick.

Issue #1024, opened 2024-03-18
`ollama suffix_url` | TypeError: 'async for' requires an object with __aiter__ method, got bytes


QIN2DIM:
**Bug description**

OLLAMA API 错

**Bug solved method**

https://github.com/QIN2DIM/MetaGPT/commit/7b6c5efc1dd951eec061af42afb92ab6072b1592

```diff
-self.suffix_url = "/chat"
+self.suffix_url = "/api/chat"
```
**Environment information**

- LLM type and model name: `ollama`
- System version: `win11`
- Python version: `3.10`

- packages version: 230315( eb8e8460b87ccfe5c4dffee56d9e955b7f9bfab6
- installation method: git clone

**Screenshots or logs**

![图片](https://github.com/geekan/MetaGPT/assets/62018067/0924b44a-962a-430f-844f-79ad47f356a4)

QIN2DIM:
去年年底开始盲测对比不同的 Agent 项目，三四个月过去了， MetaGPT 的稳定性还是只能用“千疮百孔”来形容，但总体上 idea 不错，需要努努力👍

better629:
check the `https://github.com/geekan/MetaGPT/issues/988#issuecomment-1999001192` if you have right configuration.

PR #1025, opened 2024-03-18
fix(ollama): TypeError

QIN2DIM:
Ollama | TypeError: 'async for' requires an object with __aiter__ method, got bytes #1024

```yaml
llm:
  api_type: ollama
  model: qwen:72b-chat
  base_url: http://192.168.1.180:11434
  api_key: x
```


better629:
we have given the example of `base_url: http://xxxx:11434/api` in the document.

PR #1026, opened 2024-03-18
Add Inference of SWE-bench for Data Interpreter

mannaandpoem:
**Features**
Add Inference of SWE-bench for Data Interpreter

### DIFF
--- a/data/inference/run_api.py
+++ b/data/inference/run_api.py
@@ -106,6 +106,7 @@ async def main(
         model_nickname = Path(model_name_or_path).name
     output_file = f"{model_nickname}__{dataset_name_or_path.split('/')[-1]}__{split}"
     output_file = Path(output_dir, output_file + ".jsonl")
+    output_file.parent.mkdir(parents=True, exist_ok=True)
     logger.info(f"Will write to {output_file}")
     existing_ids = set()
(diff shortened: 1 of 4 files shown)


data/inference/run_api.py
```diff
@@ -0,0 +1,159 @@
...
+    user_message = inputs.split("\n", 1)[1]
+    try:
+        await interpreter.run([system_messages, user_message])
+        code = interpreter.execute_code.nb.cells[-1].source
+        return code
+    except Exception as e:
+        logger.error(e)
+        logger.error(f"Inputs: {inputs}")
+        traceback.print_exc()
+        raise e
```
geekan, line 45: 1. I think most of this function should be written inside the interpreter
2. The two lines of logger.error should be reduced to one line

data/inference/run_api.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def gpt_tokenize(string: str, encoding) -> int:
+    """Returns the number of tokens in a text string."""
+    num_tokens = len(encoding.encode(string))
+    return num_tokens
```
geekan, line 51: This function already has a corresponding function inside metagpt and should be removed.

geekan (approved):

PR #1049, opened 2024-03-19
fix text ut error

shenchucheng:
**Features**

- Fix the unittest error in tests/metagpt/utils/test_text.py caused by the max token change of the gpt-3.5-turbo model.
- Prevent excessive output if the test_text.py unittest fails.

**Result**
image


garylin2099:
corresponding tests passed, LGTM

Issue #1034, opened 2024-03-19
Can I register my own tool, using @register_tool? I think so, but it is not working. 


krish240574:
**Bug description**
I've defined my own class, and registered it as a tool, using the @register_tool decorator. However, di does not use the tool. 

**Auxillary information**
I've even tried to tell metagpt to use the particular tool, by name in the prompt, still it does not see it. 

**Environment information**

- LLM type and model name: got-4-turbo-preview
- System version: Ubuntu 20.04
- Python version: 3.10

- packages version:
- installation method: pip install metagpt

**Screenshots or logs**

iorisa:
There are some conditions need to met before di using the tool:
1. Proper comments for the tool. Here is a demo:
```python
    async def create_repo(self, idea:str) -> Path:
        """
        Creates a new software repository with user requirements.

        Args:
            idea (str): User requirement.

        Returns:
            Path: The path to the newly created software repository.


        Example:
            To create a software repository for "snake game":

            ```python
            todo = PrepareDocuments()
            repo_path = await todo.create_repo("snake game")
            print("Repository created at:", repo_path)
            ```
        """
        self._override_context()
        await self.run(with_messages=[Message(content=idea)])
        return self.context.repo.workdir
```
The google docstring is used as the tool description during the di searches for a tool.

2. The task description needs to match the tool description. Here is a demo:
```python
import fire

from metagpt.roles.di.data_interpreter import DataInterpreter


async def main():
    prompt = f"""This is a software requirement:\n```text\nwrite a snake game\n```\n .
        1. Create a software repository.\n
        2. Write a PRD.\n
        3. Write a design.\n
        Note: All required dependencies and environments have been fully installed and configured."""
    di = DataInterpreter(tools=["GPTvGenerator", "PrepareDocuments", "WritePRD", "WriteDesign"])

    await di.run(prompt)


if __name__ == "__main__":
    fire.Fire(main)
```

Currently, the tool search ability is not perfect, if you want to test the process, you can specify the description of the tool to use in the prompt

krish240574:
@iorisa Thank you very much for your response. 
I'm guessing the run() call is local to your code, not absolutely necessary(not all examples implement the run()) - do correct me if I'm wrong. 
I've added docstring, and made sure the task description includes a clear indication about the tool and the method to be used. I've indicated tools=[<toolname>] - my specific toolname. 
Still no luck, I can't get it ro see the tool at all.

iorisa:
[...]

PR #1071, opened 2024-03-22
add custom tool example

geekan:
**Features**

- add a example of DataInterpreter for using custom tool.
    
**Feature Docs**

**Influence**

Solve https://github.com/geekan/MetaGPT/issues/1034

**Result**

image

**Other**

### DIFF
--- a/examples/di/custom_tool.py
+++ b/examples/di/custom_tool.py
@@ -27,7 +27,7 @@ def magic_function(arg1: str, arg2: int) -> dict:
 
 async def main():
     di = DataInterpreter(tools=["magic_function"])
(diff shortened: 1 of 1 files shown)


examples/di/custom_tool.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Returns:
+        dict: ...
+    """
+    return {"arg1": arg1 * 3, "arg2": arg2 * 5}
+
+
+async def main():
+    di = DataInterpreter(tools=["magic_function"])
+    await di.run("Just call the magic function with arg1 'A' and arg2 'B'. Tell me the result.")
```
garylin2099, line 30: since arg2 is int, should arg2 to be changed to, say, 1?

geekan: image

PR #1079, opened 2024-03-22
fixed regexp to add support of unicode for strings

igeni:
**Features**

- fixed regexp to add support of unicode for strings
    
**Feature Docs**

**Influence**

**Result**

**Other**
fixed regexp to add support of unicode for strings and for improving of readability


geekan (approved):
LGTM

Issue #1076, opened 2024-03-22
AttributeError: 'HumanProvider' object has no attribute 'config'


Beichen-Ma:
**Bug description**

When I try to use human engagement by running script example/build_customized_multi_agents.py with command python3 examples/build_customized_multi_agents.py --add_human True, got below error: AttributeError: 'HumanProvider' object has no attribute 'config"

**Bug solved method**

**Environment information**

MacOS 14.4, conda python 3.10, OpenAI gpt-4-turbo-preview

- LLM type and model name:
- System version:
- Python version:

- packages version:
- installation method: cd ./MetaGPT and pip install -e .

**Screenshots or logs**

<img width="1717" alt="image" src="https://github.com/geekan/MetaGPT/assets/90386322/efe4ef91-8184-4c2d-b244-930766d5190a">

geekan:
@Beichen-Ma Can you try again?

Beichen-Ma:
That works! Thanks very much.

PR #1080, opened 2024-03-23
fixbug: #1076

iorisa:
**Features**
fixbug: #1076


geekan (approved):
This implementation may still be flawed because it directly overrides the underlying method, which is not recommended behavior.

PR #1090, opened 2024-03-24
updata di example and change the location of the .ipynb locations

luxiangtaoya:
**Features**

- xx
- yy
    
**Feature Docs**

**Influence**

**Result**

**Other**


geekan (approved):
lgtm

PR #1097, opened 2024-03-25
Enable registering and retrieving tools from a path

garylin2099:
- Example usage
```python
from metagpt.tools.tool_recommend import BM25ToolRecommender

tools = ["workspace/scikit_learn/sklearn/linear_model"]
req = "Classify wine dataset with logistic regression"
tr = BM25ToolRecommender(tools=tools)

results = await tr.recall_tools(req)  # recall only

results = await tr.recommend_tools(req)  # recall + llm rank

# results are list[Tool]
print(results[0].name)
print(results[0].path)
print(results[0].code)
```
- output
Screen Shot 2024-03-25 at 4 22 37 PM


geekan (commented):
Are the test cases enough?

metagpt/tools/tool_convert.py
```diff
@@ -62,3 +81,67 @@ def get_class_method_docstring(cls, method_name):
...
+            ):
+                func_schemas = self._get_function_schemas(body_node)
+                class_schemas["methods"].update({body_node.name: func_schemas})
+        class_schemas["code"] = ast.get_source_segment(self.source_code, node)
+        self.tool_schemas[node.name] = class_schemas
+
+    def visit_FunctionDef(self, node):
+        self._visit_function(node)
+
+    def visit_AsyncFunctionDef(self, node):
```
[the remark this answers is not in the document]
garylin2099: we need this function to visit any standalone async functions in the code text

garylin2099:
> Are the test cases enough?

Not enough yet. Since its target is not main, I suggest merging it first if needed. Tests must be supplemented if merge is filed to main later.

metagpt/tools/tool_registry.py
```diff
@@ -117,25 +122,56 @@ def make_schema(tool_source_object, include, path):
...
             logger.warning(f"invalid tool name or tool type name: {key}, skipped")
     return valid_tools
+
+
+def register_tools_from_file(file_path) -> dict[str, Tool]:
+    file_name = Path(file_path).name
+    if not file_name.endswith(".py") or file_name == "setup.py" or file_name.startswith("test"):
+        return {}
+    registered_tools = {}
+    code = Path(file_path).read_text(encoding="utf-8")
```
iorisa, line 154: use `aread` to auto adapt file charset

metagpt/tools/tool_registry.py
```diff
@@ -117,25 +122,56 @@ def make_schema(tool_source_object, include, path):
...
+            schemas=schemas,
+            tool_code=tool_code,
+        )
+        registered_tools.update({name: TOOL_REGISTRY.get_tool(name)})
+    return registered_tools
+
+
+def register_tools_from_path(path) -> dict[str, Tool]:
+    tools_registered = {}
+    if os.path.isfile(path):
```
iorisa, line 170: Add a filter to make "*.py" only.

garylin2099: this filter is at register_tools_from_file

metagpt/tools/tool_registry.py
geekan: pycharm? Press shift twice to open the global symbol table and search for aread, or cmd+shift+f to open full-text search and search for aread

geekan:
I recommend using the example mentioned in the issue as a test case. In addition, try to add as few self-consistent examples as possible to illustrate the function itself.

BTW, is "workspace/scikit_learn/sklearn/linear_model" the only way to use it? `import sklearn.linear_model` and use it may be better?

metagpt/tools/tool_registry.py
garylin2099: I tried aread. However:

1. It looks unnecessary to use async operation here
2. Using it requires the function and all the related higher-level function to switch to async, which is a lot of change

One can consider providing a read_text() utils in the future for automatic encoding handling. For now, I think path.read_text is fine.

geekan (approved):

Issue #1094, opened 2024-03-25
docs/prd/xxxxxx.json not found


quarechen:
**Bug description**

Detected source code "docs/prd/20240325023557.json"

**Environment information**

use the example of docker

- LLM type and model name:  gpt-3.5-turbo or GLM-4
- System version: docker ： metagpt/metagpt latest  dcf87e6f98b9ec799bac44bc8c8ff7786fea94bb337d2fbb426da1a7cdbe6470
- Python version: Python 3.9.17

- packages version:
- installation method: 

**Screenshots or logs**

![image](https://github.com/geekan/MetaGPT/assets/30918026/252ec681-63ed-4f93-a59a-97ae6b4564fc)

iorisa:
Same issue regarding exception recovery as #1013: the recovery process didn't restore the context.repo when it was recovering.

PR #1103, opened 2024-03-25
fixbug: #1094

iorisa:
**Features**
fixbug: #1094

### DIFF
--- a/metagpt/context.py
+++ b/metagpt/context.py
@@ -112,6 +112,7 @@ class Context(BaseModel):
         return {
             "workdir": str(self.repo.workdir) if self.repo else "",
             "kwargs": {k: v for k, v in self.kwargs.__dict__.items()},
+            "cost_manager": self.cost_manager.model_dump_json(),
         }
 
     def deserialize(self, serialized_data: Dict[str, Any]):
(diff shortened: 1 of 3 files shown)


metagpt/context.py
```diff
@@ -108,3 +102,38 @@ def llm_with_cost_manager_from_llm_config(self, llm_config: LLMConfig) -> BaseLL
         if llm.cost_manager is None:
             llm.cost_manager = self._select_costmanager(llm_config)
         return llm
+
+    def serialize(self) -> Dict[str, Any]:
```
better629, line 106: Will it better to use `model_serializer`? so we can just use pydantic default methods like `model_dump`.
Same with `deserialize`.

iorisa: The naming of `serialize` is aligned with the naming in other modules.
In the `Context`, there are some objects that do not support serialization and deserialization, such as `GitRepository` and `Path`.
Moreover, the entire `Context` essentially consists of three pieces of information: the repo workdir, kwargs, and cost manager. Therefore, serializing these three is sufficient.

geekan (approved):

Issue #1100, opened 2024-03-25
debate example fail to work with gemini


iitrsamrat:
**Bug description**
debate example throws error with gemini-pro 1.5.
Websearch works with gemini-pro

**Bug solved method**

**Environment information**
Python 3.9
Conda

- LLM type and model name: Gemini-Pro
- System version:
- Python version: 3.9

**Screenshots or logs**
python3 debate.py "Talk about Artificial General Intelligence"
2024-03-25 17:57:01.666 | INFO     | metagpt.const:get_metagpt_package_root:29 - Package root set to /Users/samsaha2
2024-03-25 17:57:03.800 | INFO     | metagpt.team:invest:90 - Investment: $3.0.
2024-03-25 17:57:03.801 | INFO     | __main__:_act:63 - Biden(Democrat): to do SpeakAloud(SpeakAloud)
2024-03-25 17:57:06.072 | WARNING  | metagpt.utils.common:wrapper:572 - There is a exception in role's execution, in order to resume, we delete the newest role communication message in the role's memory.
2024-03-25 17:57:06.081 | ERROR    | metagpt.utils.common:wrapper:554 - Exception occurs, start to serialize the project, exp:
Traceback (most recent call last):
  File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/metagpt/utils/common.py", line 563, in wrapper
    return await func(self, *args, **kwargs)
  File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/metagpt/roles/role.py", line 558, in run
    rsp = await self.react()
ValueError: The `response.text` quick accessor only works for simple (single-`Part`) text responses. This response is not simple text.Use the `result.parts` accessor or the full `result.candidates[index].content.parts` lookup instead.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/metagpt/utils/common.py", line 549, in wrapper
    result = await func(self, *args, **kwargs)
  File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/metagpt/team.py", line 134, in run
    await self.env.run()
Exception: Traceback (most recent call last):
  File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/metagpt/utils/common.py", line 563, in wrapper
    return await func(self, *args, **kwargs)
[...]

PR #1105, opened 2024-03-25
fixbug: #1100

iorisa:
**Features**
fixbug: #1100

### DIFF
--- a/metagpt/provider/google_gemini_api.py
+++ b/metagpt/provider/google_gemini_api.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # @Desc   : Google Gemini LLM from https://ai.google.dev/tutorials/python_quickstart
-
+import json
 import os
-from typing import Optional, Union
+from dataclasses import asdict
+from typing import List, Optional, Union
 
 import google.generativeai as genai
 from google.ai import generativelanguage as glm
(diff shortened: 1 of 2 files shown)


metagpt/provider/google_gemini_api.py
```diff
@@ -141,7 +143,11 @@ async def _achat_completion_stream(self, messages: list[dict], timeout: int = US
         )
         collected_content = []
         async for chunk in resp:
-            content = chunk.text
+            try:
+                content = chunk.text
+            except Exception as e:
+                logger.warning(f"messages: {messages}\nerrors: {e}\n{BlockedPromptException(str(chunk))}")
+                raise BlockedPromptException(str(chunk))
```
geekan, line 150: Wouldn't it naturally throw this exception? I wonder what it would throw out naturally?

geekan (commented):
https://stackoverflow.com/questions/77723993/gemini-pro-api-blocking-replies Referring to this, is it possible to directly optimize the code?

metagpt/provider/google_gemini_api.py
iorisa: "The exception message has nothing to do with `BlockedPromptException`, it needs to be wrapped:
```text
File "/Users/samsaha2/miniconda3/envs/metagpt/lib/python3.9/site-packages/google/generativeai/types/generation_types.py", line 328, in text
raise ValueError(
ValueError: The response.text quick accessor only works for simple (single-Part) text responses. This response is not simple text.Use the result.parts accessor or the full result.candidates[index].content.parts lookup instead.
```

geekan (approved):

PR #1053, opened 2024-03-20
feat: Add software team tool lib for DI

iorisa:
**Features**
Add software team tool lib for DI

PR #1112, opened 2024-03-26
feat: +intent detect

iorisa:
**Features**
1. Detect the user intent for each word.
2. Create clarification if need.
3. Bind SOP if there is.

DEPs: 
- #1125 
- #1053 
- #1068


metagpt/actions/intent_detect.py
```diff
@@ -0,0 +1,307 @@
...
+    Attributes:
+        ref (str): The reference to the original words.
+        clarification (str): A question for the user to clarify the intention of the unclear words.
+    """
+
+    ref: str
+    clarification: str
+
+
+class _IntentDetectDialogIntentions(BaseModel):
```
stellaHSR, line 80: maybe "IntentDetectDialogIntentions" is better

iorisa: done. It has been moved into `IntentDetect`

geekan (commented):
1. Use `IntentDetection` instead of `IntentDetect`
2. This implementation lacks proof that it helps. There need to be corresponding examples to prove that it is helpful to the effect

PR #1118, opened 2024-03-27
feat(core): Add stream data return and reception

2495165664:
1. add file: utils/steam_pipe.py
2. add demo: samples/flask_web_api.py
3. Other core code modifications, Add and use the StreamPipe class at night
4. Add flask library to requirements

**Features**

- xx
- yy
    
**Feature Docs**

**Influence**
Not affecting other functions

**Result**

4b5a6673dfb9d9fe06dd0272ff89d39
Download this MD file here
e9207c3f6e6c4bf6edb7b95736221a7

**Other**

### DIFF
--- a/examples/flask_web_api.py
+++ b/examples/flask_web_api.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+@Time    : 2024/3/27 9:44
+@Author  : leiwu30
+@File    : flask_web_api.py
+@Description    : Stream log information and communicate over the network via web api.
+"""
+import os
+import json
+import socket
+import asyncio
+import threading
+
+from metagpt.utils.stream_pipe import StreamPipe
+from metagpt.roles.tutorial_assistant import TutorialAssistant
+from metagpt.const import METAGPT_ROOT
+
+from flask import Flask, Response
+from flask import request, jsonify, send_from_directory
+
+app = Flask(__name__)
+
+
+def write_tutorial(message):
+    async def main(idea, stream_pipe):
+        role = TutorialAssistant(stream_pipe=stream_pipe)
+        await role.run(idea)
+
+    def thread_run(idea: str, stream_pipe: StreamPipe = None):
+        """
+            Convert asynchronous function to thread function
+        """
+        asyncio.run(main(idea, stream_pipe))
+
+    stream_pipe = StreamPipe()
+    thread = threading.Thread(target=thread_run, args=(message["content"], stream_pipe,))
+    thread.start()
+
+    while not stream_pipe.finish:
+        stream_pipe.wait()
+        msg = stream_pipe.get_message()
+        yield stream_pipe.msg2stream(msg)
+
+    # 文件位置
+    md_file = stream_pipe.get_k_message("file_name")
+
+    yield stream_pipe.msg2stream(
+        f"\n\n[{os.path.basename(md_file)}](http://{server_address}:{server_port}/download/{md_file})")
+
+
+@app.route('/v1/chat/completions', methods=['POST'])
+def completions():
+    """
+    data: {
+        "model": "write_tutorial",
+        "stream": true,
+        "messages": [
+            {
+                "role": "user",
+                "content": "Write a tutorial about MySQL"
+            }
+        ]
+    }
+    """
+
+    data = json.loads(request.data)
+    print(json.dumps(data, indent=4, ensure_ascii=False))
+
+    # Non-streaming interfaces are not supported yet
+    stream_type = True if "stream" in data.keys() and data["stream"] else False
+    if not stream_type:
+        return jsonify({"status": 200})
+
+    # Only accept the last user information
+    last_message = data["messages"][-1]
+    model = data["model"]
+
+    # write_tutorial
+    if model == "write_tutorial":
+        return Response(write_tutorial(last_message), mimetype="text/plain")
+    else:
+        return jsonify({"status": 200})
+    # return Response(event_stream(), mimetype="text/plain")
+
+
+@app.route('/download/<path:filename>')
+def download_file(filename):
+    return send_from_directory(METAGPT_ROOT, filename, as_attachment=True)
+
+
+if __name__ == "__main__":
+    """
+    curl https://$server_address:$server_port/v1/chat/completions -X POST -d '{
+        "model": "gpt-3.5-turbo",
+        "stream": true,
+        "messages": [
+          {
+               "role": "user",
+               "content": "Write a tutorial about MySQL"
+          }
+        ]
+    }'
+    """
+    server_port = 7860
+    server_address = socket.gethostbyname(socket.gethostname())
+
+    app.run(port=server_port, host=server_address)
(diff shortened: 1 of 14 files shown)


geekan:
Awesome! Let’s start Review now!

metagpt/provider/anthropic_api.py
```diff
@@ -62,6 +62,9 @@ async def _achat_completion_stream(self, messages: list[dict], timeout: int = US
             elif event_type == "content_block_delta":
                 content = event.delta.text
                 log_llm_stream(content)
+                if self.stream_pipe:
+                    self.stream_pipe.set_message(content)
```
shenchucheng, line 66: The functionality provided by this code `self.stream_pipe.set_message(content)` duplicates that of `log_llm_stream(content)`. You can use the `set_llm_stream_logfunc` method to make `log_llm_stream` call your `stream_pipe.set_message` method.

requirements.txt
```diff
@@ -69,4 +69,5 @@ imap_tools==1.5.0  # Used by metagpt/tools/libs/email_login.py
 qianfan==0.3.2
 dashscope==1.14.1
 rank-bm25==0.2.2  # for tool recommendation
-jieba==0.42.1  # for tool recommendation
\ No newline at end of file
+jieba==0.42.1  # for tool recommendation
+flask   # for web api
```
shenchucheng, line 73: Make `flask` an extra dependency since it's not mandatory, See https://github.com/geekan/MetaGPT/blob/main/setup.py#L26.

shenchucheng (changes requested):

metagpt/provider/anthropic_api.py
2495165664: Because `log_llm_stream` uses global logs, it is considered here that web requests will generate log concurrency, resulting in web api data anomalies, so self.stream_pipe is used to ensure that a single request ensures that stream data is not interleaved. And I saw that the `set_llm_stream_logfunc method` is also not suitable for setting global log items.

shenchucheng:
@2495165664 . This is a good example, but the introduction of `stream_pipe` has intruded upon the original implementation. We can optimize it. With the help of `metagpt.logs.set_llm_stream_logfunc` and `contextvars.ContextVar`, we can achieve the same functionality without modifying any code under the `metagpt` directory.

Update the `metagpt/utils/stream_pipe.py`:

```python
...
from contextvars import ContextVar
from metagpt.logs import set_llm_stream_logfunc


stream_pipe_var: ContextVar[StreamPipe] = ContextVar("stream_pipe")


def stream_pipe_log(content):
    stream_pipe_var.get().set_message(content)


set_llm_stream_logfunc(stream_pipe_log)


def write_tutorial(message):
    async def main(idea, stream_pipe):
        stream_pipe_var.set(stream_pipe)
        role = TutorialAssistant()
        await role.run(idea)
    ...
```.

File: metagpt/provider/anthropic_api.py (hunk 1 above)
Comment (shenchucheng) [reply metagpt/provider/anthropic_api.py:66]: Using context variables (contextvar) can address the concerns you have regarding this issue. Context variables in Python allow you to pass context information in asynchronous code, which is particularly useful for resolving certain concurrency and thread safety issues.

Comment (2495165664) [reply metagpt/provider/anthropic_api.py:66]: Wow, you are right, let me rephrase it

File: examples/flask_web_api.py [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+            {
+                "role": "user",
+                "content": "Write a tutorial about MySQL"
+            }
+        ]
+    }
+    """
+
+    data = json.loads(request.data)
+    print(json.dumps(data, indent=4, ensure_ascii=False))
```
better629, line 77: use logger

examples/flask_web_api.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # Non-streaming interfaces are not supported yet
+    stream_type = True if "stream" in data.keys() and data["stream"] else False
```
better629, line 80: True if data.get("stream") else False

examples/flask_web_api.py
```diff
@@ -0,0 +1,119 @@
...
+    if not stream_type:
+        return jsonify({"status": 200})
+
+    # Only accept the last user information
+    last_message = data["messages"][-1]
+    model = data["model"]
+
+    # write_tutorial
+    if model == "write_tutorial":
```
better629, line 89: if you use a openai compatible api, `model` is usually a LLM name not `write_tutorial`, maybe you can add extra field like `feature` or others.

examples/flask_web_api.py
```diff
@@ -0,0 +1,119 @@
...
+    }
+    """
+
+    data = json.loads(request.data)
+    print(json.dumps(data, indent=4, ensure_ascii=False))
+
+    # Non-streaming interfaces are not supported yet
+    stream_type = True if "stream" in data.keys() and data["stream"] else False
+    if not stream_type:
+        return jsonify({"status": 200})
```
better629, line 82: if not support stream=False, use a status=400(bad request) or add extra msg to distinguish normal response. 
same with model != `write_tutorial`.

examples/flask_web_api.py
```diff
@@ -0,0 +1,119 @@
...
+    if model == "write_tutorial":
+        return Response(write_tutorial(last_message), mimetype="text/plain")
+    else:
+        return jsonify({"status": 200})
+    # return Response(event_stream(), mimetype="text/plain")
+
+
+@app.route('/download/<path:filename>')
+def download_file(filename):
+    return send_from_directory(METAGPT_ROOT, filename, as_attachment=True)
```
better629, line 98: seems the tutorial md output is under `TUTORIAL_PATH` ?

examples/flask_web_api.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+@Time    : 2024/3/27 9:44
+@Author  : leiwu30
+@File    : flask_web_api.py
```
better629, line 6: would filename `stream_output_via_api.py` have more clear and direct meaning?

geekan:
Hello, there are still some CRs that need to be modified. After modification, they can be merged.

examples/flask_web_api.py
2495165664: I hope to be able to seamlessly connect to `openai applications`. Currently, there is no web page for intelligent `agents` on the market. Here, the model can be used to directly access the deployed `openai applications`.
I can add a compatibility note here:
`openai['model'] ~ MetaGPT['agent']`

better629:
lgtm

better629:
@2495165664 there is a conflict
```
<<<<<<< main
    "web-api": ["flask==3.0.2"],
=======
    "android_assistant": ["pyshine==0.0.9", "opencv-python==4.6.0.66"],
>>>>>>> main
```
you should keep `web-api` and `android_assistant` both.

geekan (approved):
Although I feel like this approach is a little too complex, I still like the functionality it offers. LGTM

kiedeng:
@2495165664 @shenchucheng Due to the presence of human intervention in metagpt, how can it be paused during the streaming print process on the web end to receive input from humans?

Issue #1109, opened 2024-03-26
llm stream_response alway been print regardless of debug flag


azurewtl:
**Feature description**
`log_llm_stream` is been use for all llm_provider, and they all call the following `log_llm_stream` function
stream response with async function often result in messed up logging.
```python
logger = define_log_level()

def log_llm_stream(msg):
    _llm_stream_log(msg)

def set_llm_stream_logfunc(func):
    global _llm_stream_log
    _llm_stream_log = func

_llm_stream_log = partial(print, end="")
```

**Your Feature**
currently `loguru` doesn't provide a stable/legit way to get `log_level`, so whenever we use `define_log_level`, we need a global variable to keep track of log_level, and mute log_llm_stream corresponding. 
If that's a feasible way to do so, I could create a PR

shenchucheng:
@azurewtl Thank you for bringing up this issue. I'd like to clarify your suggestion: Are you proposing to add a new log_level configuration and treat llm stream output as debug level output?

azurewtl:
@shenchucheng yes!

shenchucheng:
@azurewtl It sounds pretty good, we'd love to see your PR.

PR #1124, opened 2024-03-27
Feature mute stream log for info level

azurewtl:
**Features**
see issue : #1109 
llm stream_response alway been print regardless of debug flag
I modified `logs.py` rather than `config2.py`.
This is because log is been set via `define_log_level` rather than `config.yaml`
To keep things simple, a global variable `_print_level` is introduced in `logs.py`
    
**Influence**
`define_log_level(print_level="INFO")` will mute `log_llm_stream`

### DIFF
--- a/metagpt/logs.py
+++ b/metagpt/logs.py
@@ -14,9 +14,13 @@ from loguru import logger as _logger
 
 from metagpt.const import METAGPT_ROOT
 
+_print_level = "INFO"
 
 def define_log_level(print_level="INFO", logfile_level="DEBUG", name: str = None):
     """Adjust the log level to above level"""
+    global _print_level
+    _print_level = print_level
+
     current_date = datetime.now()
     formatted_date = current_date.strftime("%Y%m%d")
     log_name = f"{name}_{formatted_date}" if name else formatted_date  # name a log with prefix name
@@ -31,7 +35,8 @@ logger = define_log_level()
 
 
 def log_llm_stream(msg):
-    _llm_stream_log(msg)
+  if _print_level in ["DEBUG"]:
+        _llm_stream_log(msg)
 
 
 def set_llm_stream_logfunc(func):
--- a/metagpt/logs.py
+++ b/metagpt/logs.py
@@ -35,8 +35,7 @@ logger = define_log_level()
 
 
 def log_llm_stream(msg):
-  if _print_level in ["DEBUG"]:
-        _llm_stream_log(msg)
+    _llm_stream_log(msg)
 
 
 def set_llm_stream_logfunc(func):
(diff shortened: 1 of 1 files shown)


metagpt/logs.py
```diff
@@ -31,7 +35,8 @@ def define_log_level(print_level="INFO", logfile_level="DEBUG", name: str = None
 
 
 def log_llm_stream(msg):
-    _llm_stream_log(msg)
+  if _print_level in ["DEBUG"]:
+        _llm_stream_log(msg)
```
shenchucheng, line 39: It's recommended to add this condition inside the `_llm_stream_log` function. Users can utilize `set_llm_stream_logfunc` to customize the behavior of their own `llm_stream_log` , but they may not want it to be affected by the `log_level` flag. You can define the function like this:

```python

def _llm_stream_log(msg):
    if _print_level in ["DEBUG"]:
        print(msg, end="")
```

shenchucheng (changes requested):

metagpt/logs.py
azurewtl: reasonable consideration！ change is made accordingly

shenchucheng:
@azurewtl Thanks for your improvements. It works when I call `define_log_level(print_level="DEBUG")` in my Python script entry if I want to print the LLM stream log. However, I'm considering whether to provide a simpler way to set it up, perhaps by passing environment variables `METAGPT_LOG_PRINT_LEVEL` and `METAGPT_LOG_FILE_LEVEL`?

azurewtl:
> @azurewtl Thanks for your improvements. It works when I call `define_log_level(print_level="DEBUG")` in my Python script entry if I want to print the LLM stream log. However, I'm considering whether to provide a simpler way to set it up, perhaps by passing environment variables `METAGPT_LOG_PRINT_LEVEL` and `METAGPT_LOG_FILE_LEVEL`?

I thought about using config. however, my proposal is that debug flag should be pass via command line tools.
Oftentimes, If we want to see debug info during development, I run program and pass debug flag such as:
```
python run.py --debug
``` 
Maybe YAML better off storage config that is NOT changed on ad-hoc basis. Otherwise if I want to show less or more info, I need to change config frequently.

geekan:
@shenchucheng I think this is a new requirement and can be discussed additionally. This PR can be merged first

geekan (approved):

metagpt/logs.py
```diff
@@ -14,9 +14,13 @@
 
 from metagpt.const import METAGPT_ROOT
 
+_print_level = "INFO"
```
geekan, line 17: 默认的 _print_level 应该使用 DEBUG？

geekan:
Anyway, I’ll merge first

PR #1099, opened 2024-03-25
fixbug: #1095

iorisa:
**Features**
fixbug: #1095

PR #1055, opened 2024-03-20
feat: +learn readme

iorisa:
**Features**
+ README learning action

PR #1125, opened 2024-03-27
Feature/import repo

iorisa:
**Features**
The Import process allows external existing projects to continue incrementally adding requirements.
Deps:
- #1053 
- #1055 
- #1068
- #1099 
- #1103


geekan (changes requested):
This PR has quite a few changes mixed in. It is necessary to state in the PR document what functions these modifications provide, and give corresponding examples (screenshots of successful runs). If there are multiple functions, you need to give multiple screenshots of successful operations.

iorisa:
Previously submitted features were all designed to support this feature, so it merges them.

geekan:
I still think some examples should be provided. Also, the description of this PR "The Import process allows external existing projects to continue incrementally adding requirements." does not seem to be correct. I don't seem to see the possibility that this process will work

iorisa:
You're right, there are still the following issues with the importing process:
1. Requirement merging. Corresponding to the PRD's "Requirement Pool" — the software summary in the readme is not equivalent to requirements; generating "Requirement Pool" by summary leads to misinterpretation of requirements and incorrect code changes. So, this issue remains to be resolved.
2. Design filtering. New requirements need filtering out areas that won't be changed. The current process still involves rewriting the entire codebase.
These functionalities rely on the bug fixing and new requirement evaluation processes, as well as the optimization based on their effectiveness.
Currently, only the dependencies between various stages have been resolved. While this makes the process executable, it is not yet perfect.

PR #1156, opened 2024-04-02
update:change text_to_image default size 1024x1024

Justin-ZL:
**Features**

- xx
- yy
    
**Feature Docs**

change text_to_image default size 1024x1024

**Influence**

openai text to image

**Result**

**Other**


voidking:
LGTM

PR #1161, opened 2024-04-08
Code interpreter

luxiangtaoya:
**Features**

- xx
- yy
    
**Feature Docs**

**Influence**

**Result**

**Other**


examples/rag_pipeline.py
```diff
@@ -0,0 +1,248 @@
+"""RAG pipeline"""
+
+import asyncio
+
+from pydantic import BaseModel
+
+from metagpt.const import DATA_PATH, EXAMPLE_DATA_PATH
```
better629, line 7: You should merge or rebase from the existed code branch, not just add the files as new files. The commit history will break.

luxiangtaoya:
add rag and experience block

PR #1172, opened 2024-04-10
Make RAG embedding configurable and add gpt-4-turbo in token_counter.

seehi:
**Features**

- Make RAG embedding configurable and add common embedding types such as openai/azure/gemini/ollama. Previously default using either openai or azure defined in llm_config, other embedding types require the input of embed_model.
- Add gpt-4-turbo in token_counter.
    
**Feature Docs**

**Influence**

**Result**

image

**Other**


geekan (approved):
lgtm

Issue #1239, opened 2024-04-30
RAG faiss AssertionError


AprilCat:
**Bug description**

execute this demo
```Python
import asyncio

from metagpt.rag.engines import SimpleEngine
from metagpt.rag.schema import FAISSRetrieverConfig
from metagpt.const import EXAMPLE_DATA_PATH

DOC_PATH = EXAMPLE_DATA_PATH / "rag/travel.txt"

async def main():
    engine = SimpleEngine.from_docs(input_files=[DOC_PATH], retriever_configs=[FAISSRetrieverConfig()])

    answer = await engine.aquery("What does Bob like?")
    print(answer)

if __name__ == "__main__":
    asyncio.run(main())

```
get error
```bash
Traceback (most recent call last):
  File "/home/wanfu/projects/llm/multi_agent_rag/src/simple_custom_object.py", line 26, in <module>
    asyncio.run(main())
  File "/home/wanfu/data/miniconda3/envs/metagpt/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/wanfu/data/miniconda3/envs/metagpt/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/wanfu/projects/llm/multi_agent_rag/src/simple_custom_object.py", line 21, in main
    engine.add_docs([DOC_PATH])
  File "/mnt/data/work/development/projects/llm/MetaGPT/metagpt/rag/engines/simple.py", line 195, in add_docs
    self._save_nodes(nodes)
  File "/mnt/data/work/development/projects/llm/MetaGPT/metagpt/rag/engines/simple.py", line 274, in _save_nodes
    self.retriever.add_nodes(nodes)
  File "/mnt/data/work/development/projects/llm/MetaGPT/metagpt/rag/retrievers/faiss_retriever.py", line 12, in add_nodes
    self._index.insert_nodes(nodes, **kwargs)
  File "/home/wanfu/data/miniconda3/envs/metagpt/lib/python3.9/site-packages/llama_index/core/indices/vector_store/base.py", line 320, in insert_nodes
    self._insert(nodes, **insert_kwargs)
  File "/home/wanfu/data/miniconda3/envs/metagpt/lib/python3.9/site-packages/llama_index/core/indices/vector_store/base.py", line 311, in _insert
    self._add_nodes_to_index(self._index_struct, nodes, **insert_kwargs)
[...]

Issue #1213, opened 2024-04-21
RAG Faiss AssertionError


GG-Lizen:
**Environment information**

- LLM type and model name: ollama ,nomic-embed-text
- System version:win 11 
- Python version:3.9
- MetaGPT version or branch:0.8

**Bug description**

run code as below
```
import asyncio
from metagpt.rag.engines import SimpleEngine
from metagpt.rag.schema import FAISSRetrieverConfig
from metagpt.const import EXAMPLE_DATA_PATH
DOC_PATH = EXAMPLE_DATA_PATH / "rag/travel.txt"
async def main():
    engine = SimpleEngine.from_docs(input_files=[DOC_PATH], retriever_configs=[FAISSRetrieverConfig()])
    answer = await engine.aquery("What does Bob like?")
    print(answer)

if __name__ == "__main__":
    asyncio.run(main())
```
occur AssertionError
```
Traceback (most recent call last):
  File "E:\MyTask\Metagpt\MetaGPT-0.8.0\examples\rag_test.py", line 25, in <module>
    asyncio.run(main())
  File "D:\Dev_Software\Anaconda\envs\metagpt\lib\asyncio\runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "D:\Dev_Software\Anaconda\envs\metagpt\lib\asyncio\base_events.py", line 649, in run_until_complete
    return future.result()
  File "E:\MyTask\Metagpt\MetaGPT-0.8.0\examples\rag_test.py", line 15, in main
    SimpleEngine.from_docs(input_files=[DOC_PATH], retriever_configs=retriever_configs).persist(persist_dir)
  File "e:\mytask\metagpt\metagpt-0.8.0\metagpt\rag\engines\simple.py", line 111, in from_docs
    return cls._from_index(index, llm=llm, retriever_configs=retriever_configs, ranker_configs=ranker_configs)
  File "e:\mytask\metagpt\metagpt-0.8.0\metagpt\rag\engines\simple.py", line 211, in _from_index
    retriever = get_retriever(configs=retriever_configs, index=index)  # Default index.as_retriever
  File "e:\mytask\metagpt\metagpt-0.8.0\metagpt\rag\factories\retriever.py", line 52, in get_retriever
    retrievers = super().get_instances(configs, **kwargs)
  File "e:\mytask\metagpt\metagpt-0.8.0\metagpt\rag\factories\base.py", line 18, in get_instances
[...]

PR #1241, opened 2024-05-01
Make RAG embedding model dimension configurable

usamimeri:
## **User description**
For #1213 and #1239. When using  text-embedding-3-large(the dimension is 3072) from OpenAI, or some other embedding models deployed by fastchat(e.g. bce-embedding-base_v1 the dimension is 512), it will cause assertion error due to mismatch between query and index.

Consequently,I think it's reasonable to make dimension configurable.

```yaml
# fastchat API
embedding:
  api_type: "openai"
  base_url: "http://127.0.0.1:8000/v1/embeddings"
  model: "bce-embedding-base_v1"
  dimensions: 512
```

### DIFF
--- a/config/config2.example.yaml
+++ b/config/config2.example.yaml
@@ -18,6 +18,7 @@ embedding:
   model: ""
   api_version: ""
   embed_batch_size: 100
+  dimensions: 
 
 repair_llm_output: true  # when the output is not a valid json, try to repair it
 
--- a/config/config2.example.yaml
+++ b/config/config2.example.yaml
@@ -18,7 +18,7 @@ embedding:
   model: ""
   api_version: ""
   embed_batch_size: 100
-  dimensions: 
+  dimensions: # output dimension of embedding model
 
 repair_llm_output: true  # when the output is not a valid json, try to repair it
 
--- a/metagpt/rag/schema.py
+++ b/metagpt/rag/schema.py
@@ -48,7 +48,7 @@ class FAISSRetrieverConfig(IndexRetrieverConfig):
             self.dimensions = config.embedding.dimensions or self._embedding_type_to_dimensions.get(
                 config.embedding.api_type, 1536
             )
-            if config.embedding.api_type not in self._embedding_type_to_dimensions:
+            if not config.embedding.dimensions and config.embedding.api_type not in self._embedding_type_to_dimensions:
                 logger.warning(
                     f"You didn't set dimensions in config when using {config.embedding.api_type}, default to 1536"
                 )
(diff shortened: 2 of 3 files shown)


config/config2.example.yaml
```diff
@@ -18,6 +18,7 @@ embedding:
   model: ""
   api_version: ""
   embed_batch_size: 100
+  dimensions:
```
geekan, line 21: Can you add comments?

metagpt/rag/schema.py
```diff
@@ -34,16 +35,21 @@ class IndexRetrieverConfig(BaseRetrieverConfig):
...
     """Config for FAISS-based retrievers."""
 
-    dimensions: int = Field(default=0, description="Dimensionality of the vectors for FAISS index construction.")
+    dimensions: int = Field(
+        default=config.embedding.dimensions, description="Dimensionality of the vectors for FAISS index construction."
+    )
 
     _embedding_type_to_dimensions: ClassVar[dict[EmbeddingType, int]] = {
         EmbeddingType.GEMINI: 768,
-        EmbeddingType.OLLAMA: 4096,
```
geekan, line 41: Is this an expected removal?

usamimeri: Yes. as https://github.com/ollama/ollama/issues/651 mentioned, the output dimension of ollama embedding may not be 4096.

metagpt/rag/schema.py
```diff
@@ -34,16 +35,21 @@ class IndexRetrieverConfig(BaseRetrieverConfig):
...
     }
 
     @model_validator(mode="after")
     def check_dimensions(self):
-        if self.dimensions == 0:
+        if self.dimensions is None:
+            if config.embedding.api_type not in self._embedding_type_to_dimensions:
+                logger.info(
```
seehi, line 50: Prefer warning

metagpt/rag/schema.py
usamimeri: Thank you for pointing that out, indeed need to consider backward compatibility.

metagpt/rag/schema.py
```diff
@@ -44,7 +45,13 @@ class FAISSRetrieverConfig(IndexRetrieverConfig):
     @model_validator(mode="after")
     def check_dimensions(self):
         if self.dimensions == 0:
-            self.dimensions = self._embedding_type_to_dimensions.get(config.embedding.api_type, 1536)
+            self.dimensions = config.embedding.dimensions or self._embedding_type_to_dimensions.get(
+                config.embedding.api_type, 1536
+            )
+            if config.embedding.api_type not in self._embedding_type_to_dimensions:
```
seehi, line 51:
```python
if not config.embedding.dimensions and ...
```

seehi:
LGTM

geekan (approved):

PR #1268, opened 2024-05-13
fix openai image schema bug

usamimeri:
### **User description**
I try to run android assistant with gpt-4-turbo, but there was an error indicating the messages schema was incorrect:
`"Invalid type for 'messages[1].content[1].image_url': expected an object, but got a string instead`
by checking the openai-sdk doc https://platform.openai.com/docs/guides/vision  and https://platform.openai.com/docs/api-reference/chat/create
"I found that the format {"type":"image_url","image_url":YOUR_IMAGE_URL} can be used on gpt-4-vision-preview, but not on gpt-4-turbo, whereas the format {"type":"image_url","image_url":{"url":YOUR_IMAGE_URL}} can be used on both."

Consequently, I update the schema for more compability.


better629:
lgtm

PR #1295, opened 2024-05-24
 Fix OllamaEmbedding

xunzhimeng:
**Features**

- Fix OllamaEmbedding parameter error, OllamaEmbedding  need model_name but give model.
    
**Feature Docs**

**Influence**

**Result**

**Other**


seehi:
FYI: https://github.com/geekan/MetaGPT/pull/1299

xunzhimeng:
> FYI: #1299

I got it, in some case, like AzureOpenAIEmbedding,there no model_name parameter.Maybe you should check it.

Issue #1297, opened 2024-05-24
ReAct loop can not stop gracefully, usually not stop.


Wei-Jianan:
**Bug description**
1. I created a role that simply counts up to 3, after which the goal is considered accomplished. All the LLM I tried usually would not set next_state to -1 in role._think()  when operating in ReAct mode.
2. After I essentially solved 1.  the role encountered difficulties in exiting gracefully. 
```
2024-05-24 08:33:16.872 | INFO     | metagpt.roles.role:_think:397 - End actions with next_state=-1
2024-05-24 08:33:16.874 | WARNING  | metagpt.utils.common:wrapper:649 - There is a exception in role's execution, in order to resume, we delete the newest role communication message in the role's memory.
Traceback (most recent call last):
  File "/home/wei/kf/MetaGPT/metagpt/utils/common.py", line 640, in wrapper
    return await func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wei/kf/MetaGPT/metagpt/roles/role.py", line 551, in run
    rsp = await self.react()
          ^^^^^^^^^^^^^^^^^^
  File "/home/wei/kf/MetaGPT/metagpt/roles/role.py", line 520, in react
    rsp = await self._react()
          ^^^^^^^^^^^^^^^^^^^
  File "/home/wei/kf/MetaGPT/metagpt/roles/role.py", line 475, in _react
    rsp = await self._act()
          ^^^^^^^^^^^^^^^^^
  File "/home/wei/kf/MetaGPT/workspace/counter.py", line 53, in _act
    logger.info(f"{self._setting}: to do {self.rc.todo}({self.rc.todo.name})")
                                                         ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'name'

During handling of the above exception, another exception occurred:
```
**Bug solved method**
1. ![1716540205523](https://github.com/geekan/MetaGPT/assets/32776997/8eee4c5c-bae3-4c24-b259-6045c298ed39)
I suppose this issue arises because the LLM might get confused after being instructed 'Just answer a number between 0-{n_states}', and then being told 'If you think you have completed your goal and don't need to go to any of the stages, return -1.' LLM may not be able to return -1. 
2.
```
class Role(SerializationMixin, ContextMixin, BaseModel):
    async def _think(self) -> bool:
        """Consider what to do and decide on the next course of action. Return false if nothing can be done."""
[...]

Issue #1260, opened 2024-05-10
react模式下，自动拼装的promt，缺少action描述信息，请问如何解决？


nzk1912:
如下图中，只有类名，请问如何增加工具的功能描述，并且能够拼接到promt中，以便更好的选择正确的action？
![image](https://github.com/geekan/MetaGPT/assets/16470935/d63dee05-4f74-471d-86f5-0fe89457cbb8)

YangZeyu95:
Same issue.
Now I have a workaround, set 'desc' when calling set_actions().
For example, self.set_actions([action_dummy(desc='some description')])

hs3180:
You can override __str__ method in your actions. They will be showed in action ReAct prompt.

https://github.com/geekan/MetaGPT/blob/7057ace34468f36f28f8948455e7d251a3ed2899/metagpt/roles/role.py#L279

PR #1298, opened 2024-05-24
[fix] 1. react loop not exit 2. Action.desc in role.states

Wei-Jianan:
**Features**
-- https://github.com/geekan/MetaGPT/issues/1297
-- https://github.com/geekan/MetaGPT/issues/1260
  

**Influence**
-- It might affect all the Role that run in ReAct mode.


geekan:
This issue seems to have been discussed many times, but the current tendency is not to modify the 0/-1 part for the time being. If there is any doubt about action desc, we can do additional discussion

PR #1306, opened 2024-05-28
fix openai_api.py usage bug

usamimeri:


usamimeri:
duplicate https://github.com/geekan/MetaGPT/pull/1294

PR #1307, opened 2024-05-29
Add swe_agent

mannaandpoem:
**Features**
Add swe_agent


metagpt/const.py
```diff
@@ -132,3 +132,6 @@ def get_metagpt_root():
 GENERALIZATION = "Generalize"
 COMPOSITION = "Composite"
 AGGREGATION = "Aggregate"
+
+# Reporter
+METAGPT_REPORTER_DEFAULT_URL = os.environ.get("METAGPT_REPORTER_URL", "")
```
geekan, line 137: Variable parts should be placed in config, not const, unless it is a path

metagpt/tools/swe_agent_commands/_setup_default_env.sh
```diff
@@ -0,0 +1,20 @@
+# _setup_default_env.sh
+# Default Mode from SWE-Bench
+# https://github.com/princeton-nlp/SWE-agent/blob/ca54d5556b9db4f4f2be21f09530ce69a72c0305/config/configs/default_sys-env_window100-detailed_cmd_format-last_5_history-1_demos.yaml
```
geekan, line 3: Please check the corresponding license and indicate whether it involves license declaration issues. If so, please declare it in the appropriate place.

metagpt/utils/git_repository.py
```diff
@@ -32,6 +46,18 @@ class ChangeType(Enum):
     UNTRACTED = "U"  # File is untracked (not added to version control)
 
 
+class RateLimitError(Exception):
+    def __init__(self, message="Rate limit exceeded"):
+        self.message = message
+        super().__init__(self.message)
```
geekan, line 52: Should this class be placed here?

PR #1311, opened 2024-05-29
LLMConfig add seed

chenchun0629:
Features: LLMConfig add seed

### DIFF
--- a/metagpt/configs/llm_config.py
+++ b/metagpt/configs/llm_config.py
@@ -5,6 +5,7 @@
 @Author  : alexanderwu
 @File    : llm_config.py
 """
+
 from enum import Enum
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


metagpt/configs/llm_config.py
```diff
@@ -76,6 +77,7 @@ class LLMConfig(YamlModel):
     best_of: Optional[int] = None
     n: Optional[int] = None
     stream: bool = True
+    seed: Optional[int] = None
```
geekan, line 80: If you just add the seed here but don't use it in the code, it won't take effect.

geekan (approved):
Then add the underlying logic of seed

geekan:
注意到底层逻辑没有实现，这里单独增加 seed 是没用的。如果你有时间，希望你可以补齐底层逻辑

Issue #1304, opened 2024-05-27
Possible bug with the Role Memory


nkrj01:
**Bug description**
When I put two agents (AgentA and AgentB) in an environment, AgentB is automatically picks up the published message of AgentA even though _watch function is not set to AgentAAction. 

AgentAAction: take a string and publish one letter at a time
AgentBAction: given any string simple add the word "Add"

There is no LLM functionality here.

Code is shown below:

 ## Defining AgentA
from metagpt.actions import Action, UserRequirement
from metagpt.roles import Role
from metagpt.logs import logger
from metagpt.schema import Message

class AgentAAction(Action):
    name: str = "AgentAAction"

    async def run(self, task: str):
        sub_task = [letter for letter in task]
        return sub_task

class AgentA(Role):
    name: str = "AgentA"
    profile: str = "AgentA"

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.set_actions([AgentAAction])
        self._watch({UserRequirement})

    async def _act(self) -> None:
        logger.info(f"{self._setting}: to do {self.rc.todo}({self.rc.todo.name}) {type(self.rc)}")
        todo = self.rc.todo  # todo will be SimpleWriteCode()

        task = self.get_memories(k=1)[0]  # find the most recent messages
        sub_tasks = await todo.run(task.content)
        
        for i in sub_tasks:
           msg = Message(content=i, role=self.profile, cause_by=type(todo))
           self.rc.env.publish_message(msg)

## Defining AgentB
class AgentBAction(Action):
    name: str = "AgentBAction"

    async def run(self, sub_task: str):
        final_output = sub_task + "Add"
        return final_output

class AgentB(Role):
    name: str = "AgentB"
    profile: str = "AgentB"

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.set_actions([AgentBAction])
        self._watch({UserRequirement}) ## Note!! Not watching AgentAAction

    async def _act(self) -> Message:
        logger.info(f"{self._setting}: to do {self.rc.todo}({self.rc.todo.name}) {type(self.rc)}")
        todo = self.rc.todo  # todo will be AgentBAction

        sub_tasks = self.rc.history  # find all the messages
        print(sub_tasks)
        
        for i in sub_tasks:
            final_output = await todo.run(i.content)
            final_output = Message(content=final_output, role=self.profile, cause_by=type(todo))
            self.rc.env.publish_message(final_output)
[...]

PR #1314, opened 2024-05-31
fix: Update in environment publish message

Kaushal-26:
- Publish message in env. also considering cause_by, to only those roles having cause_by in watch action of interest.

**Features**

- Now only the roles, whose `watch actions` are set same as `metagpt.schema.message.cause_by` will get the message.

**Result**

- Resolves this open issue: https://github.com/geekan/MetaGPT/issues/1304

### DIFF
--- a/metagpt/environment/base_env.py
+++ b/metagpt/environment/base_env.py
@@ -185,7 +185,7 @@ class Environment(ExtEnv):
         found = False
         # According to the routing feature plan in Chapter 2.2.3.2 of RFC 113
         for role, addrs in self.member_addrs.items():
-            if is_send_to(message, addrs):
+            if is_send_to(message, addrs) and role.is_watch(message.cause_by):
                 role.put_message(message)
                 found = True
         if not found:


metagpt/environment/base_env.py
```diff
@@ -185,7 +185,7 @@ def publish_message(self, message: Message, peekable: bool = True) -> bool:
         found = False
         # According to the routing feature plan in Chapter 2.2.3.2 of RFC 113
         for role, addrs in self.member_addrs.items():
-            if is_send_to(message, addrs):
+            if is_send_to(message, addrs) and role.is_watch(message.cause_by):
```
iorisa, line 188: - Only SOPs(standard operating procedure, like software team) use `cause_by`; 
- If it is a DI (`Data Interpreter`) or a `Generic Action`(have not implement yet), it can handle any message sent to it.
So this modification is not appropriate.

Kaushal-26: Ok fair enough, If it breaks something then it is irrelevant to change.

Somethings which motivated me to do this changes:
- Lots of irrelevant message in `role.rc.msg_buffer`.
- Most of the time `role._observe` filters according to `cause_by` As here: https://github.com/geekan/MetaGPT/blob/bc1cd11a4b284e509056cc0052ec52d20568a4ab/metagpt/roles/role.py#L432

Don't get me wrong but I want to be clear on my side:
- I get it that you guys are using all messages in `Data Interpreter`. But you guys aren't using `Environment` in [Data Interpreter Role](https://github.com/geekan/MetaGPT/blob/main/metagpt/roles/di/data_interpreter.py) and I could not find anywhere in example of DI: https://github.com/geekan/MetaGPT/tree/main/examples/di

Kaushal-26: Closing this pull, as this breaks Werewolf game: https://github.com/geekan/MetaGPT/tree/main/examples/werewolf_game

iorisa: The concept of "environment" is only needed when multiple agents coexist. The environment is responsible for establishing communication channels for them. 

The modules of MetaGPT can be adapted to different granularities of use cases. If you only need to tune the LLM, you can use the classes in the `metagpt/provider` folder; if you need a single agent, create a role object; only when you need multiple agents to collaborate do you need to create an "Environment" object, allowing the multiple agents to communicate with each other.

Issue #1313, opened 2024-05-30
AttributeError: 'Message' object has no attribute 'restricted_to'


lizhongv:
![image](https://github.com/geekan/MetaGPT/assets/49021250/36174081-2e1d-40ee-be58-0994948b33b2)

```python
 python start_game.py
2024-05-31 02:42:33.932 | INFO     | metagpt.const:get_metagpt_package_root:29 - Package root set to /data0/lizhong/MetaGPT
2024-05-31 02:42:39.139 | INFO     | __main__:start_game:34 - Game setup:
Player1: Villager,
Player2: Werewolf,
Player3: Seer,
Player4: Guard,
Player5: Villager,
Player6: Werewolf,
Player7: Witch,
2024-05-31 02:42:39.165 | INFO     | metagpt.team:invest:93 - Investment: $20.0.
2024-05-31 02:42:39.172 | ERROR    | metagpt.utils.common:wrapper:631 - Exception occurs, start to serialize the project, exp:
Traceback (most recent call last):
  File "/data0/lizhong/MetaGPT/metagpt/utils/common.py", line 640, in wrapper
    return await func(self, *args, **kwargs)
  File "/data0/lizhong/MetaGPT/metagpt/roles/role.py", line 545, in run
    if not await self._observe():
AttributeError: 'Message' object has no attribute 'restricted_to'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data0/lizhong/MetaGPT/metagpt/utils/common.py", line 626, in wrapper
    result = await func(self, *args, **kwargs)
  File "/data0/lizhong/MetaGPT/metagpt/team.py", line 131, in run
    await self.env.run()
Exception: Traceback (most recent call last):
  File "/data0/lizhong/MetaGPT/metagpt/utils/common.py", line 640, in wrapper
    return await func(self, *args, **kwargs)
  File "/data0/lizhong/MetaGPT/metagpt/roles/role.py", line 545, in run
    if not await self._observe():
  File "/data0/lizhong/MetaGPT/metagpt/ext/werewolf/roles/base_player.py", line 63, in _observe
    if len(m.restricted_to) and self.profile not in m.restricted_to and self.name not in m.restricted_to:
  File "/data0/lizhong/anaconda3/envs/cu117-pt2.0/lib/python3.10/site-packages/pydantic/main.py", line 811, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
[...]

PR #1316, opened 2024-06-02
fix: Werewolf Game

Kaushal-26:
- As all the roles in werewolf game as `react_mode` = `RoleReactMode.REACT`, so `role._react()` will be called everytime.
- When not returning anything from `role._think()` => Therefore, `role._act()` will not be executed.
- Due to which this issue was comming: https://github.com/geekan/MetaGPT/issues/1313

### DIFF
--- a/metagpt/ext/werewolf/roles/base_player.py
+++ b/metagpt/ext/werewolf/roles/base_player.py
@@ -86,7 +86,7 @@ class BasePlayer(Role):
             # FIXME: hard code to split, restricted为"Moderator"或"Moderator, 角色profile"
             # Moderator加密发给自己的，意味着要执行角色的特殊动作
             self.rc.todo = self.special_actions[0]()
-        return self.rc.todo is not None
+        return True
 
     async def _act(self):
         # todo为_think时确定的，有两种情况，Speak或Protect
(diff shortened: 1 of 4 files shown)


metagpt/ext/werewolf/roles/base_player.py
```diff
@@ -86,6 +86,7 @@ async def _think(self):
             # FIXME: hard code to split, restricted为"Moderator"或"Moderator, 角色profile"
             # Moderator加密发给自己的，意味着要执行角色的特殊动作
             self.rc.todo = self.special_actions[0]()
+        return self.rc.todo is not None
```
garylin2099, line 89: I recommend directly returning True for easier understanding. In the werewolf case, we always want to trigger the following _act once we enter _think

garylin2099 (commented):
Thanks for the fix! Commented.

metagpt/ext/werewolf/roles/base_player.py
Kaushal-26: > I recommend directly returning True for easier understanding. In the werewolf case, we always want to trigger the following _act once we enter _think

Ok, Changing this for all files.

Issue #1322, opened 2024-06-05
Unable to run metagpt


yousia33:
(metagpt) PS F:\codetest\Metagpt> metagpt "create a 2048 game"
2024-06-05 16:46:27.754 | INFO     | metagpt.const:get_metagpt_package_root:29 - Package root set to f:\codetest\metagpt
╭─────────────────────── Traceback (most recent call last) ────────────────────────╮
│ f:\codetest\metagpt\metagpt\software_company.py:108 in startup                   │
│                                                                                  │
│   105 │   │   typer.echo("Missing argument 'IDEA'. Run 'metagpt --help' for more │
│   106 │   │   raise typer.Exit()                                                 │
│   107 │                                                                          │
│ ❱ 108 │   return generate_repo(                                                  │
│   109 │   │   idea,                                                              │
│   110 │   │   investment,                                                        │
│   111 │   │   n_round,                                                           │
│                                                                                  │
│ f:\codetest\metagpt\metagpt\software_company.py:30 in generate_repo              │
│                                                                                  │
│    27 │   recover_path=None,                                                     │
│    28 ) -> ProjectRepo:                                                          │
│    29 │   """Run the startup logic. Can be called from CLI or other Python scrip │
│ ❱  30 │   from metagpt.config2 import config                                     │
│    31 │   from metagpt.context import Context                                    │
│    32 │   from metagpt.roles import (                                            │
│    33 │   │   Architect,                                                         │
│                                                                                  │
│ f:\codetest\metagpt\metagpt\config2.py:164 in <module>                           │
│                                                                                  │
│   161 │   return result                                                          │
│   162                                                                            │
│   163                                                                            │
│ ❱ 164 config = Config.default()                                                  │
│   165                                                                            │
│                                                                                  │
│ f:\codetest\metagpt\metagpt\config2.py:106 in default                            │
│                                                                                  │
│   103 │   │   dicts = [dict(os.environ)]                                         │
│   104 │   │   dicts += [Config.read_yaml(path) for path in default_config_paths] │
│   105 │   │   final = merge_dict(dicts)                                          │
│ ❱ 106 │   │   return Config(**final)                                             │
│   107 │                                                                          │
│   108 │   @classmethod                                                           │
│   109 │   def from_llm_config(cls, llm_config: dict):                            │
│                                                                                  │
│ F:\anacoda\envs\metagpt\Lib\site-packages\pydantic\main.py:176 in __init__       │
│                                                                                  │
[...]

Issue #1290, opened 2024-05-22
Validation Error 


cyborg3poch:
ValidationError: 1 validation error for Config llm   Field required [type=missing, input_value={'PATH': '/Users/psyb0rg/..._INIT_AT_FORK': 'FALSE'}, input_type=dict]     For further information visit https://errors.pydantic.dev/2.7/v/missing

usamimeri:
can you provide the content of your config2.yaml? The error message indicates that the llm field is missing from config2.yaml which should be:

```yaml
llm:
  api_type:…
  api_key:…
```

usamimeri:
> > Can you provide the content of your config2.yaml? The error message indicates that the llm field is missing from config2.yaml which should be:
> > ```yaml
> > llm:
> >   api_type:…
> >   api_key:…
> > ```
> 
> I use Gemini and this is my content but it's still have validation error
> 
> ```
> llm:
>   api_type: "gemini"
>   model: "gemini-pro"
>   api_key: "My_API_Key"
> ```

It works for me. Try to pull the latest code to ensure everything is up-to-date.

garylin2099:
> ValidationError: 1 validation error for Config llm Field required [type=missing, input_value={'PATH': '/Users/psyb0rg/..._INIT_AT_FORK': 'FALSE'}, input_type=dict] For further information visit https://errors.pydantic.dev/2.7/v/missing

Make sure you have updated ~/.metagpt/config2.yaml or config/config2.yaml (for detail please see [here](https://docs.deepwisdom.ai/main/en/guide/get_started/configuration/llm_api_configuration.html)). Can you check if the problem still exists?

Issue #1220, opened 2024-04-23
pydantic_core._pydantic_core.ValidationError: 1 validation error for Config


1027548366:
D:\MetaGPT-main\venv\Scripts\python.exe D:/MetaGPT-main/test.py
2024-04-23 17:14:39.661 | INFO     | metagpt.const:get_metagpt_package_root:29 - Package root set to D:\MetaGPT-main
Traceback (most recent call last):
  File "D:\MetaGPT-main\test.py", line 3, in <module>
    repo: ProjectRepo = generate_repo("Create a 2048 game")  # or ProjectRepo("<path>")
  File "D:\MetaGPT-main\metagpt\software_company.py", line 30, in generate_repo
    from metagpt.config2 import config
  File "D:\MetaGPT-main\metagpt\config2.py", line 164, in <module>
    config = Config.default()
  File "D:\MetaGPT-main\metagpt\config2.py", line 106, in default
    return Config(**final)
  File "D:\MetaGPT-main\venv\lib\site-packages\pydantic\main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Config
llm.api_key
  Value error, Please set your API key in config2.yaml [type=value_error, input_value='YOUR_API_KEY', input_type=str]
    For further information visit https://errors.pydantic.dev/2.5/v/value_error

I've written the config as to why this is still the case？？？
![微信图片_20240423172525](https://github.com/geekan/MetaGPT/assets/91006305/804b3f17-98b4-4137-9487-475ead58560b)

seehi:
Try to create .metagpt in the user's root directory, such as `C:\Users\x\.metagpt\config2.yaml`.

PR #1324, opened 2024-06-05
optimize error message when validating api_key failed

usamimeri:
**Features**
Added a more user-friendly error message for API_KEY validation failure for https://github.com/geekan/MetaGPT/issues/1290, https://github.com/geekan/MetaGPT/issues/1322 and  https://github.com/geekan/MetaGPT/issues/1220.
Users can use `metagpt --init-config` to add a config2.yaml under `home/username/.metagpt/config2.yaml`. 
However, the config file in the home directory will overwrite the one in the MetaGPT repository, which may cause unexpected results. Therefore, the error message has been optimized.
**Result**
1. When two paths exist or only the config under home path exists and there is no api_key in the home path one:
```shell
ValidationError: 1 validation error for Config
llm.api_key
  Value error, Please set your API key in /home/usamimeri/.metagpt/config2.yaml. If you also set your config in /home/usamimeri/MetaGPT/config/config2.yaml, 
the former will overwrite the latter. This may cause unexpected result.
 [type=value_error, input_value='YOUR_API_KEY', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error
```
2. When only the config under the repository exists:
```shell
ValidationError: 1 validation error for Config
llm.api_key
  Value error, Please set your API key in /home/usamimeri/MetaGPT/config/config2.yaml [type=value_error, input_value='YOUR_API_KEY', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error
```

### DIFF
--- a/metagpt/configs/llm_config.py
+++ b/metagpt/configs/llm_config.py
@@ -94,8 +94,8 @@ class LLMConfig(YamlModel):
     @classmethod
     def check_llm_key(cls, v):
         if v in ["", None, "YOUR_API_KEY"]:
-            repo_config_path=(METAGPT_ROOT / "config/config2.yaml")
-            root_config_path=(CONFIG_ROOT / "config2.yaml")
+            repo_config_path = METAGPT_ROOT / "config/config2.yaml"
+            root_config_path = CONFIG_ROOT / "config2.yaml"
             if root_config_path.exists():
                  raise ValueError(
                     f"Please set your API key in {root_config_path}. If you also set your config in {repo_config_path}, \nthe former will overwrite the latter. This may cause unexpected result.\n")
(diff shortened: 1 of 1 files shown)


metagpt/configs/llm_config.py
```diff
@@ -94,7 +94,15 @@ class LLMConfig(YamlModel):
     @classmethod
     def check_llm_key(cls, v):
         if v in ["", None, "YOUR_API_KEY"]:
-            raise ValueError("Please set your API key in config2.yaml")
+            repo_config_path=(METAGPT_ROOT / "config/config2.yaml")
```
better629, line 97: no need add (), xx_path = METAGPT_ROOT xxx

usamimeri: resolved

PR #1343, opened 2024-06-13
bug fix: list index out of range

weinaike:
**Features**


metagpt/roles/role.py
```diff
@@ -374,7 +374,7 @@ async def _think(self) -> bool:
             if self.rc.max_react_loop != len(self.actions):
                 self.rc.max_react_loop = len(self.actions)
             self._set_state(self.rc.state + 1)
-            return self.rc.state >= 0 and self.rc.state < len(self.actions)
+            return (self.rc.state + 1) >= 0 and (self.rc.state + 1) < len(self.actions)
```
garylin2099, line 377: `self._set_state` has incremented self.rc.state and therefore no need for another increment

Issue #1356, opened 2024-06-19
Research role fails if some of Serpapi searches return no results


igor-pechersky:
**Bug description**
Research role fails if some of Serpapi searches return no results

**Reproduction**
***Configuration***
```
llm:
  api_type: "openai"  
  model: "gpt-4o"  
  base_url: "https://api.openai.com/v1"  # or forward url / other llm url
  api_key: "sk-XXX"

search:
  api_type: 'serpapi' 
  api_key: 'XXX'
  params:
    tbs: qdr:y #last year
    gl: us
    hl: en
```
***Command line***
```
python3 -m metagpt.roles.researcher 'Coulianu Tozgrec'
```
**Bug solved method**
TBD

**Environment information**
- LLM type and model name: gpt-4o
- System version: Linux codespaces-27135d 6.5.0-1021-azure 
- Python version: Python 3.9.17
- MetaGPT version or branch: main, after https://github.com/geekan/MetaGPT/commit/38cea1daf2b87ebc31a56c995d7857d07289fa70

**Screenshots or logs**
Traceback (most recent call last):
  File "/workspaces/MetaGPT/metagpt/utils/common.py", line 640, in wrapper
    return await func(self, *args, **kwargs)
  File "/workspaces/MetaGPT/metagpt/roles/role.py", line 550, in run
    rsp = await self.react()
  File "/workspaces/MetaGPT/metagpt/roles/researcher.py", line 96, in react
    msg = await super().react()
  File "/workspaces/MetaGPT/metagpt/roles/role.py", line 519, in react
    rsp = await self._react()
  File "/workspaces/MetaGPT/metagpt/roles/role.py", line 474, in _react
    rsp = await self._act()
  File "/workspaces/MetaGPT/metagpt/roles/researcher.py", line 63, in _act
    summaries = await asyncio.gather(*todos)
  File "/workspaces/MetaGPT/metagpt/roles/researcher.py", line 61, in <genexpr>
    todos = (todo.run(*url, query=query, system_text=research_system_text) for (query, url) in links.items())
TypeError: run() missing 1 required positional argument: 'url'

shenchucheng:
Thank you for your PR https://github.com/geekan/MetaGPT/pull/1357 to fix this issue. Now that it has been merged, we will be closing this issue.

PR #1357, opened 2024-06-19
Research role fails if some of Serpapi searches return no results

igor-pechersky:
https://github.com/geekan/MetaGPT/issues/1356


shenchucheng:
LGTM

Issue #1341, opened 2024-06-13
integration with UI frameworks...


nothk:
I want to integrate metagpt with UI framework, 
but all metagpt example and demo are under CLI...

do we have any UI integration example or demo,,,
which UI framework should I use?

thanks a lot,

voidking:
As far as I know, you can redirect the output to socket and then you can show the info as you wish. Refer to https://github.com/geekan/MetaGPT/blob/main/examples/stream_output_via_api.py

PR #1358, opened 2024-06-19
feat: Integration of MetaGPT with UI framework, Chainlit

Kaushal-26:
**Features**

- MetaGPT on UI, Integrated using Chainlit: https://github.com/Chainlit/chainlit, as it gives flexibility with python.
- Message `Streaming` working ✅
- We can send an "idea" from UI as "Create a 2048 Game".
- Added 2 chat starters, so easily clickable to run the idea, also can run on any other idea given.
- Added support for `Action Node` output. To show as `json markdown`.

**Feature Docs**

_To Setup with own application_
- We can change `Environment.run`, `Team.run`, `Role.run`, `Role._act`, `Action.run`.
- In this code, changed `Environment.run`, as it was easier to do.
- We will need to change `metagpt.logs.set_llm_stream_logfunc` to stream messages in UI with Chainlit Message.
- To use at some other place we need to call `chainlit.Message(content="").send()` with content.

**Influence**

- Beginner friendly who are comfortable in working with UI.

**Usage**
```bash
pip install chainlit
chainlit run app.py
```
- Now go to: http://localhost:8000
- Select,
  - `Create a 2048 game`
  - `Write a cli Blackjack Game`
  - `Type your own message...`
- It will run a software company.

**Result**

metagpt

**Other**

- This cannot be used directly in production, as it uses `global` message for streaming, it will require [User Session](https://docs.chainlit.io/concepts/user-session)
- resolve: https://github.com/geekan/MetaGPT/issues/1341


voidking:
LGTM
Great job!

geekan (approved):

chaos-max:
hi, if i set a role as a human(is_human=true),is seems doesn't work right, do you have a way to solve this

PR #1368, opened 2024-06-27
Fix: update regular expression of parse code

SevenFo:
**Features**

Change non-greedy matching to greedy matching prevent matching from being interrupted by another ` ``` `


geekan:
Great. I think we still need to prevent multiple ``` here? If you just change the matching range, multiple ```s are unexpected.

geekan:
Since this method does not satisfy all scenarios, it is closed.

PR #1381, opened 2024-07-04
Fixing Streamed Output Misalignment

kiedeng:
When running stream_output_via_api.py with Flask, if there are multiple concurrent requests, a situation may arise where the responses are mismatched: data meant for request A is returned to request B. The main reason for this issue is the failure to initialize member variables properly during instantiation. The instantiation occurs at: stream_pipe = StreamPipe().


shenchucheng:
LGTM

geekan (approved):

PR #1415, opened 2024-07-26
feat: +git commit history

iorisa:
**Features**
- Synchronize `git_repository.py` under GitLab.
- Synchronize `shell.py` under GitLab.
- Add `history` property to `git_repository` to support repository version switching.


geekan:
Afterwards, they will be merged together at once and will not be reviewed separately.

PR #1421, opened 2024-07-29
feat: + write code only

iorisa:
**Features**
- Added a process that directly allows `Engineer` to write code.


geekan:
Afterwards, they will be merged together at once and will not be reviewed separately.

Issue #1430, opened 2024-07-31
DO NOT FORCE VALIDATE '{'Required Python packages'}' by default


hgftrdw45ud67is8o89:
**Bug description**
`metagpt\actions\action_node.py", line 432, in _aask_v1
    instruct_content = output_class(**parsed_data)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "..........\Lib\site-packages\pydantic\main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for PM_NODE_AN
  Value error, Missing fields: {'Required Python packages'} `

**Bug solved method**
DO NOT VALIDATE THIS FIELD. user may ask the agents to do non py related stuff,why would we force this validate and introduce a hard error? Seems silly.

**Environment information**
irrelevant

- LLM type and model name:
- MetaGPT version or branch:0.8.1

**Screenshots or logs**
`action_node.py", line 432, in _aask_v1
    instruct_content = output_class(**parsed_data)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".........\Lib\site-packages\pydantic\main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for PM_NODE_AN
  Value error, Missing fields: {'Required Python packages'} [type=value_error, input_value={'Required Rust packages'...ption for backup data.'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/value_error

The above exception was the direct cause of the following exception:

Traceback (most recent call last):`

thisisvk45:
It appears that you encountered a `ValidationError` related to missing fields when using the Pydantic model in MetaGPT. Specifically, the error indicates that the `'Required Python packages'` field is missing:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for PM_NODE_AN
Value error, Missing fields: {'Required Python packages'}
```

Proposed Solution:
To address this issue, I suggest modifying the validation behavior of the `'Required Python packages'` field. Since users might request actions unrelated to Python packages, enforcing validation on this field can introduce unnecessary errors. Therefore, we should make this field optional.

 Solution Details

1. Update the Pydantic Model:
[...]

PR #1435, opened 2024-08-04
Fix bug https://github.com/geekan/MetaGPT/issues/1430 and also make some field in design api optional.

hgftrdw45ud67is8o89:
**Features**
- fix https://github.com/geekan/MetaGPT/issues/1430
- make some key in design_api_an.py optional,because from a few test run with non py projects. it seems to not have it. Could be reverted as MUST, if we can guarantee to make sure all relevant projects generate those design api.
- same pr feature as https://github.com/geekan/MetaGPT/pull/1431 but implemented using latest git.(please merge this quick for low git conflict)   

**Feature Docs**

**Influence***
Let a lot of hard error become non error and continue to *cook* project.

**Result**

**Other**

### DIFF
--- a/metagpt/actions/design_api_an.py
+++ b/metagpt/actions/design_api_an.py
@@ -5,7 +5,7 @@
 @Author  : alexanderwu
 @File    : design_api_an.py
 """
-from typing import List
+from typing import List,Optional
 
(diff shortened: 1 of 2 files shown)


iorisa:
LGTM

Issue #1429, opened 2024-07-31
there's a problem about using external LLM of huoshan


Gingerkiang:
MetaGPT is not supporting 'ark' as its api_type
<img width="948" alt="截屏2024-07-31 21 05 13" src="https://github.com/user-attachments/assets/8fbca465-307b-4a63-852d-92f74ddbf3f0">
<img width="733" alt="截屏2024-07-31 19 53 45" src="https://github.com/user-attachments/assets/57e39328-0cc2-4e1a-91b4-c9289ee65837">

iorisa:
developing

Gingerkiang:
Cool！

PR #1438, opened 2024-08-05
fixbug: there's a problem about using external LLM of huoshan #1429

iorisa:
**Features**
- #1429


metagpt/provider/ark_api.py
```diff
@@ -16,11 +37,45 @@ class ArkLLM(OpenAILLM):
     见：https://www.volcengine.com/docs/82379/1263482
     """
 
+    aclient: Optional[AsyncArk] = None
```
better629, line 40: Seems ark has a openai compatible interface, should we use the official sdk?

better629: ok, to support also ak/sk

metagpt/utils/token_counter.py
```diff
@@ -187,6 +187,14 @@
     "mixtral-8x7b": {"prompt": 0.4, "completion": 1.6},
 }
 
+# https://console.volcengine.com/ark/region:ark+cn-beijing/model
+DOUBAO_TOKEN_COSTS = {
```
better629, line 191: There also doubao-xx related token cost under `TOKEN_COSTS`, you can delete it.

iorisa: 截屏2024-08-07 15 08 09

The model name has the version number embedded in it.
`DOUBAO_TOKEN_COSTS` provides model names without version numbers for `pricing_plan` configuration.

metagpt/provider/ark_api.py
iorisa: The openai compatible interface is not working, so it's replaced by sdk.

better629:
lgtm

PR #1444, opened 2024-08-09
Update Action graph Version 0.1.2

didiforgithub:
**Features**
Update Basic Pipeline for humaneval, GSM8K & HotPotQA


geekan:
Since there is already a new version, this version will be closed first.

PR #1474, opened 2024-09-04
Adding Together API to the providers

abirharrasse:
We make it possible for users of MetaGPT to use their together api and get access to the agentic features offered by MetaGPT.

### DIFF
--- a/metagpt/__init__.py
+++ b/metagpt/__init__.py
@@ -5,3 +5,34 @@
 # @File    : __init__.py
 
 from metagpt import _compat as _  # noqa: F401
+from metagpt.provider.google_gemini_api import GeminiLLM
+from metagpt.provider.ollama_api import OllamaLLM
+from metagpt.provider.openai_api import OpenAILLM
+from metagpt.provider.zhipuai_api import ZhiPuAILLM
+from metagpt.provider.azure_openai_api import AzureOpenAILLM
+from metagpt.provider.metagpt_api import MetaGPTLLM
+from metagpt.provider.human_provider import HumanProvider
+from metagpt.provider.spark_api import SparkLLM
+from metagpt.provider.qianfan_api import QianFanLLM
+from metagpt.provider.dashscope_api import DashScopeLLM
+from metagpt.provider.anthropic_api import AnthropicLLM
+from metagpt.provider.bedrock_api import BedrockLLM
+from metagpt.provider.ark_api import ArkLLM
+
+__all__ = [
+    "GeminiLLM",
+    "OpenAILLM",
+    "ZhiPuAILLM",
+    "AzureOpenAILLM",
+    "MetaGPTLLM",
+    "OllamaLLM",
+    "HumanProvider",
+    "SparkLLM",
+    "QianFanLLM",
+    "DashScopeLLM",
+    "AnthropicLLM",
+    "BedrockLLM",
+    "ArkLLM",
+    "OwnLLM",
+    "TogetherLLM"
+]
--- a/metagpt/configs/llm_config.py
+++ b/metagpt/configs/llm_config.py
@@ -10,9 +10,9 @@ from typing import Optional
 
 from pydantic import field_validator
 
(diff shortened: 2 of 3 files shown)


metagpt/__init__.py
```diff
@@ -5,3 +5,34 @@
 # @File    : __init__.py
 
 from metagpt import _compat as _  # noqa: F401
+from metagpt.provider.google_gemini_api import GeminiLLM
```
better629, line 8: these codes are already in `metagpt/provider/__init__.py`

metagpt/configs/llm_config.py
```diff
@@ -10,9 +10,9 @@
 
 from pydantic import field_validator
 
-from metagpt.const import CONFIG_ROOT, LLM_API_TIMEOUT, METAGPT_ROOT
+from metagpt.const import LLM_API_TIMEOUT
 from metagpt.utils.yaml_model import YamlModel
-
+from metagpt.const import METAGPT_ROOT, CONFIG_ROOT
```
better629, line 15: use pre-commit to format code

metagpt/configs/llm_config.py
```diff
@@ -97,13 +97,12 @@ def check_llm_key(cls, v):
...
-                raise ValueError(
-                    f"Please set your API key in {root_config_path}. If you also set your config in {repo_config_path}, \nthe former will overwrite the latter. This may cause unexpected result.\n"
-                )
+                 raise ValueError(
+                    f"Please set your API key in {root_config_path}. If you also set your config in {repo_config_path}, \nthe former will overwrite the latter. This may cause unexpected result.\n")
             elif repo_config_path.exists():
                 raise ValueError(f"Please set your API key in {repo_config_path}")
             else:
-                raise ValueError("Please set your API key in config2.yaml")
+                raise ValueError(f"Please set your API key in config2.yaml")
```
better629, line 105: why add f?

metagpt/provider/together.py
```diff
@@ -0,0 +1,126 @@
+import asyncio
+from typing import Optional, Union, AsyncGenerator
+from together import Together
```
better629, line 3: doesn't add together version in requirments.txt

metagpt/provider/together.py
```diff
@@ -0,0 +1,126 @@
...
+        response = await asyncio.to_thread(self.client.chat.completions.create, **kwargs)
+        
+        collected_messages = []
+        current_sentence = ""
+        for chunk in response:
+            if chunk.choices:
+                chunk_message = chunk.choices[0].delta.content
+                if chunk_message:
+                    current_sentence += chunk_message
+                    if chunk_message.endswith(('.', '!', '?', '\n')):
```
better629, line 55: why deal with there special chars?

metagpt/provider/together.py
```diff
@@ -0,0 +1,126 @@
...
+                        collected_messages.append(current_sentence)
+                        yield current_sentence
+                        print(current_sentence, end='', flush=True)  # Print each sentence as it's completed
```
better629, line 58: use log_llm_stream instead of print

metagpt/provider/together.py
```diff
@@ -0,0 +1,126 @@
...
+from metagpt.configs.llm_config import LLMConfig, LLMType
+from metagpt.provider.llm_provider_registry import register_provider
+from metagpt.const import USE_CONFIG_TIMEOUT
+from metagpt.logs import logger
+from metagpt.utils.cost_manager import CostManager
+from metagpt.utils.token_counter import count_input_tokens, count_output_tokens, get_max_completion_tokens
+from metagpt.utils.exceptions import handle_exception
+
+@register_provider(LLMType.TOGETHER)
+class TogetherLLM(BaseLLM):
```
better629, line 15: Notice that together has openai api compatibility inferface `https://docs.together.ai/docs/openai-api-compatibility`. So it seems that no need to create a new provider here.

geekan:
We know that its api is compatible with the openai api, so in theory no new class is needed. But if I'm wrong, reopen it.

PR #1522, opened 2024-10-20
Enhanced few code files in the main branch.

vignesh1507:
Key changes that i have made to the code are as follows:

1. Exception Handling.
2. Deprecation Warning.
3. Logging.
4. Auto-Archiving.

This updated version includes more robust handling for running out of funds and provides better logging for project execution.

Made changes as you suggested @geekan, could you please review this PR?

### DIFF
--- a/metagpt/logs.py
+++ b/metagpt/logs.py
@@ -8,19 +8,16 @@
 
 import sys
 from datetime import datetime
+from functools import partial
+from typing import Callable, Optional
 
 from loguru import logger as _logger
 
 from metagpt.const import METAGPT_ROOT
 
-_print_level = "INFO"
 
-
-def define_log_level(print_level="INFO", logfile_level="DEBUG", name: str = None):
-    """Adjust the log level to above level"""
-    global _print_level
-    _print_level = print_level
-
+def define_log_level(print_level: str = "INFO", logfile_level: str = "DEBUG", name: Optional[str] = None) -> _logger:
+    """Adjust the log level to the specified levels."""
     current_date = datetime.now()
     formatted_date = current_date.strftime("%Y%m%d")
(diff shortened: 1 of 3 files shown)


geekan:
Since these changes involve different modules, I invite relevant developers to serve as reviewers

metagpt/software_company.py
```diff
@@ -39,9 +38,6 @@ def generate_repo(
     )
     from metagpt.team import Team
 
-    if config.agentops_api_key != "":
```
voidking, line 42: Why was agentops removed?

metagpt/logs.py
```diff
@@ -34,15 +31,15 @@ def define_log_level(print_level="INFO", logfile_level="DEBUG", name: str = None
...
 
-def set_llm_stream_logfunc(func):
+def set_llm_stream_logfunc(func: Callable[[str], None]) -> None:
+    """Set the function to be used for logging LLM streams."""
     global _llm_stream_log
     _llm_stream_log = func
 
 
-def _llm_stream_log(msg):
-    if _print_level in ["INFO"]:
```
shenchucheng, line 47: It looks like this changes have reverted the feature introduced in PR #1124, which allowed the `define_log_level` method to control stream logging via the `print_level` parameter.  Could you please clarify the reason for this change?

metagpt/team.py
```diff
@@ -130,7 +130,12 @@ async def run(self, n_round=3, idea="", send_to="", auto_archive=True):
                 logger.debug("All roles are idle.")
                 break
             n_round -= 1
-            self._check_balance()
+            try:
+                self._check_balance()
+            except NoMoneyException as e:
```
better629, line 135: already has a `serialize_decorator`, should we add this?

PR #1552, opened 2024-10-30
Fix JSON Load Error & Fix Claude Token Calculate Error.

didiforgithub:
Fix the JSON loading error in data_utils.py.
Add type hint to prevent unnecessary downloads.


geekan:
Can you explain how this modification fixes the corresponding bug?

didiforgithub:
For the current version, an error occurs in the load_results function due to reading an empty JSON. To resolve this issue, we fixed a portion of the code to ensure that the first optimization won't result in a max retry error.
img_v3_02g5_6c57f4d7-8639-49c3-9423-d75a87663e2g
Additionally, we added Claude's token count feature to accurately calculate Claude's token consumption.

geekan (approved):

PR #1376, opened 2024-07-01
Feature/multi llm

iorisa:
**Features**
- Allow the `Action` object to use the specified LLM configuration; 
- Support for multiple LLM configurations. If no LLM is specified, the default LLM configuration is used.

PR #1561, opened 2024-10-31
fix bug in multi llm

zhaoweiguo:
**Features**
fix bug of multi llm  #1376 

**Influence**
when calling
```python
self.set_actions([action1, action2, action3])
```

will cause
```python
class Role(SerializationMixin, ContextMixin, BaseModel):
    def _init_action(self, action: Action):
        if not action.private_config:
            action.set_llm(self.llm, override=True)   #  !! Attention: here will reset llm
        else:
            action.set_llm(self.llm, override=False)
        action.set_prefix(self._get_prefix())
```

### DIFF
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -49,7 +49,6 @@ class Action(SerializationMixin, ContextMixin, BaseModel):
             llm = create_llm_instance(config)
             llm.cost_manager = data.llm.cost_manager
             data.llm = llm
-            data.config = config   # if not set self.config, self.llm will be reset when you call Role.set_actions function
         return data
 
     @property
--- a/metagpt/actions/action.py
+++ b/metagpt/actions/action.py
@@ -49,6 +49,7 @@ class Action(SerializationMixin, ContextMixin, BaseModel):
             llm = create_llm_instance(config)
             llm.cost_manager = data.llm.cost_manager
             data.llm = llm
+            data.config = config   # if not set self.config, self.llm will be reset when you call Role.set_actions function
         return data
 
     @property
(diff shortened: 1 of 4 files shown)


metagpt/actions/action.py
```diff
@@ -49,6 +49,7 @@ def _update_private_llm(cls, data: Any) -> Any:
             llm = create_llm_instance(config)
             llm.cost_manager = data.llm.cost_manager
             data.llm = llm
+            data.config = config   # if not set self.config, self.llm will be reset when you call Role.set_actions function
```
iorisa, line 52: Thank you for providing the bug information.

The issue is caused by a logical error in the `Role._init_action` method in `metagpt/roles/role.py`. The `_init_action` method should check `private_llm` instead of `private_config`, because:

1. The user can set the LLM directly, in which case `private_config` is None.
2. If the user sets `private_config`, both `private_config` and `private_llm` will be set (not None).

Therefore, the solution to this problem should be to change `Role._init_action`, not the downstream logic.

```python
def _init_action(self, action: Action):
-        if not action.private_config:
+        if not action.private_llm:
             action.set_llm(self.llm, override=True)
         else:
             action.set_llm(self.llm, override=False)
```

Here is a unit test that you can use and add to `tests/metagpt/actions/test_action.py`:

```python
def test_set_llm():
    config1 = Config.default()
    config2 = Config.default()
    config2.llm.model = "deepseek-coder-mock"

    context = Context(config=config1)
    act = Action(context=context)
    assert act.config.llm.model == config1.llm.model

    llm2 = create_llm_instance(config2.llm)
    act.llm = llm2
    assert act.llm.model == llm2.model

    role = Role(context=context)
    role.set_actions([act])
    assert act.llm.model == llm2.model

    role1 = Role(context=context)
    act1 = Action(context=context)
    assert act1.config.llm.model == config1.llm.model
    act1.config = config2
    role1.set_actions([act1])
    assert act1.llm.model == llm2.model
```

geekan (approved):

PR #1562, opened 2024-10-31
增加对千帆实时搜索和模型人设的支持

zhaoweiguo:
**Features**

增加对千帆实时搜索和模型人设的支持

### DIFF
--- a/metagpt/configs/llm_config.py
+++ b/metagpt/configs/llm_config.py
@@ -94,6 +94,10 @@ class LLMConfig(YamlModel):
     # For Network
     proxy: Optional[str] = None
 
+    # For baidu qianfan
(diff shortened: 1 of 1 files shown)


metagpt/configs/llm_config.py
```diff
@@ -94,6 +94,10 @@ class LLMConfig(YamlModel):
     # For Network
     proxy: Optional[str] = None
 
+    # For baidu qianfan
+    system: bool = False       # 模型人设
+    disable_search: bool = True   # 是否关闭实时搜索
```
geekan, line 99: The modification must include the actual parameter changes of the underlying API call, otherwise it will not take effect.

geekan (commented):
btw, use **English**

PR #1563, opened 2024-10-31
chore: Added back to top , contributors section and star history graph

Ranjana761:
## Description

Added back to top , contributors section and star history graph in Readme.


voidking:
LGTM

geekan:
I've been wondering whether I should let this go.

PR #1579, opened 2024-11-06
postpone the import of external different providers

davidleon:
postpone the import of external different providers to lift the requirement forcing user installing different non-related provider apis.

**Features**
Postpone the external lib loading, and users only install required lib.

    
**Feature Docs**
NA

**Influence**
No

**Result**
works in my case.

**Other**
NA

### DIFF
--- a/metagpt/provider/__init__.py
+++ b/metagpt/provider/__init__.py
@@ -5,20 +5,36 @@
 @Author  : alexanderwu
 @File    : __init__.py
 """
+import importlib
+class LLMFactory:
+    def __init__(self, module_name, instance_name):
+        self.module_name = module_name
+        self.instance_name = instance_name
+        self._module = None
 
-from metagpt.provider.google_gemini_api import GeminiLLM
-from metagpt.provider.ollama_api import OllamaLLM
-from metagpt.provider.openai_api import OpenAILLM
-from metagpt.provider.zhipuai_api import ZhiPuAILLM
-from metagpt.provider.azure_openai_api import AzureOpenAILLM
-from metagpt.provider.metagpt_api import MetaGPTLLM
-from metagpt.provider.human_provider import HumanProvider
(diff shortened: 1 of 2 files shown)


metagpt/provider/__init__.py
```diff
@@ -5,20 +5,36 @@
...
+        return getattr(self._module, name)
+    def __instancecheck__(self, instance):
+        if self._module is None:
+            self._module = importlib.import_module(self.module_name)
+        return isinstance(instance, getattr(self._module, self.instance_name))
+
+    
+GeminiLLM = LLMFactory("metagpt.provider.google_gemini_api ", "GeminiLLM")
+OllamaLLM = LLMFactory("metagpt.provider.ollama_api ", "OllamaLLM")
+OpenAILLM = LLMFactory("metagpt.provider.openai_api ", "OpenAILLM")
```
better629, line 27: It can't support usage like below
```
from metagpt.config2 import config
from metagpt.provider import OpenAILLM

llm_config = config.llm
llm = OpenAILLM(llm_config)
```

metagpt/provider/__init__.py
```diff
@@ -5,20 +5,36 @@
...
+        if self._module is None:
+            self._module = importlib.import_module(self.module_name)
+        return getattr(self._module, name)
+    def __instancecheck__(self, instance):
+        if self._module is None:
+            self._module = importlib.import_module(self.module_name)
+        return isinstance(instance, getattr(self._module, self.instance_name))
+
+    
+GeminiLLM = LLMFactory("metagpt.provider.google_gemini_api ", "GeminiLLM")
```
better629, line 25: The module_name and provider_name are duplicated in `metagpt/provider/llm_provider_registry.py`

davidleon: now the duplication removed. and support the required usecase.

PR #1589, opened 2024-11-11
Added a callback mechanism for capturing logs and updates during DI execution. Can be extended to other roles too. 

krish240574:
**Features**

Logs capture and log update mechanism for the DataInterpreter, Role, Team and Env classes, which can get displayed on a webpage using websockets. 

- xx
- yy
    
**Feature Docs**

**Influence**

**Result**

**Other**

### DIFF
--- a/examples/gui_update_example.py
+++ b/examples/gui_update_example.py
@@ -1,0 +1,200 @@
+import asyncio
+import datetime
+from fastapi import FastAPI, WebSocket
+from fastapi.responses import HTMLResponse
+from metagpt.logs import logger
+from metagpt.actions import Action
+from metagpt.utils.websocket_client import WebSocketClient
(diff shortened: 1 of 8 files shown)


davidleon:
team.py change contains space tab formatting change. Better retain the original formatting style. and minimize formatting change. Any example of using the gui_callback available? a small example would be great to illustrate how this enriches the frontend experience.

krish240574:
Sorry about that, David, I'll change back to spaces - let me put up an example of the usage , in a bit.  Thanks for pointing that out.

krish240574:
@davidleon - have fixed the formatting in team.py and added an example inside examples/gui_update_example.py.

krish240574:
@davidleon - updated the user level example in examples/gui_update_example.py - better updates, better handling of the messages. Do take a look.

Issue #1643, opened 2024-12-16
botocore error on bedrock models that are specific to a region


TorgemanTarak:
**Bug description**
When I try to use a region gated model on  AWS bedrock i get the following error for example here with claude 3.5 sonnet v1:

`botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Invocation of model ID anthropic.claude-3-5-sonnet-20240620-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.`

**Bug solved method**
I think it is because the region is not passed properly to the session.client in line 41 of provider/beckrock_api.py
I will make a pull request to fix it
it just involves changing 
`client = session.client(service_name)`
to
`client = session.client(service_name,region_name = self.__credential_kwargs["region_name"])`
**PR link:**
https://github.com/geekan/MetaGPT/pull/1644
**Environment information**

- LLM type and model name:  anthropic.claude-3-5-sonnet-20240620-v1:0 
- Sytem version: Sonoma 14.7.1
- Python version: 3.9
- MetaGPT commit hash id: 68b7dc6c16894a7aea6eedd5a2d9b74c9d2642a0

shenchucheng:
Closed this issue because https://github.com/geekan/MetaGPT/pull/1644 has been merged."

PR #1644, opened 2024-12-16
pass the reigion name to the boto3.session.client to use region gated…

TorgemanTarak:
A small fix to be able to use aws bedrock models that are specific to some region.

**bugfix**
https://github.com/geekan/MetaGPT/issues/1643


voidking:
LGTM

shenchucheng (approved):
LGTM

PR #1695, opened 2025-02-14
update role atomization capacity example

jason-jszhang:
**Features**

- update role atomization capacity example
    
**Feature Docs**

**Influence**

**Result**

**Other**


examples/write_game_code.py
```diff
@@ -0,0 +1,24 @@
+import asyncio
```
seehi, line 1: 关于运行env的例子，可以参考tests/scripts/run_mgx_env.py

Issue #1709, opened 2025-02-21
openai o1 raise BadRequestError: Unsupported parameter: 'max_tokens' is not supported with this model


iorisa:
**Bug description**
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

llm config:
```yaml
llm:
  api_type: "openai"  # or azure / ollama / groq etc.
  model: "o1"  # or gpt-3.5-turbo
  base_url: "https://api.openai.com/v1"  # or forward url / other llm url
  api_key:
```

PR #1710, opened 2025-02-21
fixbug: #1703 & 

iorisa:
**Fixbugs**
- #1703 
- #1709


better629:
o1-mini does not currently support system_prompt. Can you check if it support temperature=0 and stream=True @iorisa

PR #1712, opened 2025-02-21
fix(provider): ensure base_url is always included in params

ElvisClaros:
Previously, the base_url was only added to params if a proxy was configured. This change ensures that base_url is included in params regardless of whether a proxy is set. This improves the flexibility and correctness of the configuration handling.


better629:
`base_url` is always updated in `_make_client_kwargs` and no need to include in the params here.

PR #1727, opened 2025-02-28
Feat merge cr

jason-jszhang:
**Features**

- xx
- yy
    
**Feature Docs**

**Influence**

**Result**

**Other**


metagpt/rag/engines/simple.py
```diff
@@ -31,7 +31,7 @@
     TransformComponent,
 )
 
-from metagpt.config2 import config
+from metagpt.config2 import Config
```
better629, line 34: why not use `https://github.com/geekan/MetaGPT/blob/main/metagpt/config2.py#L169`

PR #2067, opened 2026-06-14
Security: Fix command injection via mermaid.path config (#2037)

### LINKED ISSUE #2037: [Security] Command Injection via Mermaid `path` Configuration
## **TL;DR**

Identified a command injection issue in MetaGPT's Mermaid rendering flow.

MetaGPT allows users to configure the Mermaid CLI executable through the `mermaid.path` field in `config.yaml`. That value is treated as a raw string and is later passed into shell-based execution in multiple places.

As a result, if an attacker can tamper with the Mermaid `path` configuration, they can inject arbitrary shell syntax and execute commands in the security context of the user running MetaGPT.

For example, a payload in the form below is sufficient in principle to achieve arbitrary command execution:

```shell
; bash -c '...' #
```

This issue is caused by unsafe command construction and shell execution, not by Mermaid itself.

## **Root Cause**

The `mermaid.path` value is handled as a raw string and is later embedded into shell commands.

There are two unsafe execution points:

1. `metagpt/utils/common.py`
    - `check_cmd_exists()` directly concatenates the configured value into a shell command and executes it with `os.system()`.
2. `metagpt/utils/mermaid.py`
    - `mermaid_to_file()` builds a shell command using the configured Mermaid path and executes it with `asyncio.create_subprocess_shell()`.

Because both locations rely on shell parsing, the `path` field can contain shell metacharacters and injected commands.

## **Proof of Concept Code**

Below is a minimal PoC that triggers the Mermaid rendering path.

```yaml
llm:
  api_type:"openai"
  base_url:""
  api_key:""
  model:""
  temperature:0.0
  stream:true
  reasoning:false
  timeout:600

repair_llm_output:true
prompt_schema:"json"

workspace:
  path:"/data/agent-sec/tmp/repos/MetaGPT/workspace"

mermaid:
  engine:"nodejs"
  # path: "mmdc"
  path:"mmdc; /bin/bash -c '...' #'"
  puppeteer_config:"~/.metagpt/puppeteer-config.json"
```

```python
#!/usr/bin/env python3
import asyncio
from pathlib import Path

from metagpt.actions.design_api import WriteDesign
from metagpt.actions.write_prd import WritePRD
from metagpt.config2 import Config
from metagpt.context import Context

PROMPT = """
Generate a minimal PRD and system design for a Pomodoro task management web application.
""".strip()

async def main():
    cfg = Config.default(reload=True)
    ctx = Context(config=cfg)

    base = Path("/data/agent-sec/tmp/repos/MetaGPT/workspace/mermaid_flow")
    base.mkdir(parents=True, exist_ok=True)
    prd_path = base / "docs" / "prd.json"
    design_path = base / "docs" / "system_design.json"

    prd = WritePRD(context=ctx)
    prd_result = await prd.run(
        user_requirement=PROMPT,
        output_pathname=str(prd_path),
        extra_info="The output must reliably contain Mermaid diagram fields, and the diagram content itself must be directly renderable by Mermaid CLI.",
    )
    print(prd_result)

    design = WriteDesign(context=ctx)
    design_result = await design.run(
        user_requirement=PROMPT,
        prd_filename=str(prd_path),
        output_pathname=str(design_path),
        extra_info="To reiterate: the classDiagram and sequenceDiagram in the system design must be valid Mermaid syntax, without comments or explanatory prefixes/suffixes.",
    )
    print(design_result)

    expected = [
        base / "docs" / "prd-competitive-analysis.svg",
        base / "docs" / "system_design-class-diagram.svg",
        base / "docs" / "system_design-sequence-diagram.svg",
    ]
    for path in expected:
        print(f"{path} exists={path.exists()}")

if __name__ == "__main__":
    asyncio.run(main())
```

This code does not rely on any special Mermaid content. It only needs to reach the `nodejs` Mermaid execution path so that the configured `mermaid.path` value is used.

### **1. `metagpt/utils/common.py`**

```python
def check_cmd_exists(command) -> int:
    if platform.system().lower() == "windows":
        check_command = "where " + command
    else:
        check_command = "command -v " + command + ' >/dev/null 2>&1 || { echo >&2 "no mermaid"; exit 1; }'
    print(f"check command:{check_command}")
    result = os.system(check_command)
    return result
```

- `command` comes from configuration.
- It is concatenated directly into a shell command string.
- The code executes that string via `os.system()`.
- Any shell metacharacters inside `command` will be interpreted by the shell.

This means command injection can occur even before MetaGPT actually invokes `mmdc`.

### **2. `metagpt/utils/mermaid.py`**

```python
if config.mermaid.puppeteer_config:
    commands = [
        config.mermaid.path,
        "-p",
        config.mermaid.puppeteer_config,
        "-i",
        str(tmp),
        "-o",
        output_file,
        "-w",
        str(width),
        "-H",
        str(height),
    ]
else:
    commands = [config.mermaid.path, "-i", str(tmp), "-o", output_file, "-w", str(width), "-H", str(height)]
process = await asyncio.create_subprocess_shell(
    " ".join(commands), stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
)
```

- `config.mermaid.path` is attacker-controlled if the config can be modified.
- The command list is converted back into a single shell string with `" ".join(commands)`.
- That string is then executed with `asyncio.create_subprocess_shell()`.
- Shell syntax inside `config.mermaid.path` is interpreted as part of the command line.

In other words, this code does not treat `mermaid.path` as a path. It treats it as shell text.

## **Reproduction**

1. Set `mermaid.engine` to `nodejs`.
2. Set `mermaid.path` to an injected string, for example a reverse shell payload.
1. Trigger any code path that calls `metagpt.utils.mermaid.mermaid_to_file(...)`.
2. Observe that the injected command executes.

GouravSingal-code:
**Features**

Closes #2037. Fixes two unsafe shell-execution sites that allowed arbitrary command injection via the user-controllable `mermaid.path` configuration value.

- `metagpt/utils/common.py::check_cmd_exists` — replace `os.system()` + shell string concat with `shutil.which()`. Same return-code contract (0 if found, non-zero if not), no shell involved, cross-platform.
- `metagpt/utils/mermaid.py::mermaid_to_file` — replace `asyncio.create_subprocess_shell(" ".join(commands))` with `asyncio.create_subprocess_exec(*commands)`. argv passed directly, no shell parsing.

This incidentally also closes the same class of injection via `config.mermaid.puppeteer_config`, which was previously joined into the same shell string.

**Influence**

No behavior change for legitimate paths (`"mmdc"`, `/usr/local/bin/mmdc`, etc.). The default install / happy path is completely unaffected. Only inputs containing shell metacharacters are now rejected instead of silently executed by the shell.

Threat surface this protects against:
- Multi-tenant deployments where users supply configs
- Shared templates / examples that ship a `config2.yaml`
- Supply-chain attacks via dependency-provided configs
- CI pipelines that load untrusted fixture configs

**Result**

Added two regression tests:
- `tests/metagpt/utils/test_common.py::test_check_cmd_exists_rejects_shell_injection` — covers 5 metacharacter payloads (empty string, `;`, `&&`, `\$()`, `|`); asserts non-zero return AND that the smuggled command never executes.
- `tests/metagpt/utils/test_mermaid.py::test_mermaid_to_file_rejects_path_injection` — drives `mermaid_to_file` end-to-end with the exact payload format from issue #2037 using a fake mmdc binary in `tmp_path`; asserts no payload execution.

Verification performed locally:
- The issue's exact PoC payload (`\"mmdc; /bin/bash -c '...' #'\"`) does not execute against the patched code; the same payload against the pre-patch code triggers the embedded command.
- Happy path (legitimate `mmdc`) still runs and produces an SVG.
- All other existing tests in `test_common.py` / `test_mermaid.py` behave identically to `main` (the 3 pre-existing failures in `test_parse_data_exception[2]`, `test_parse_recipient[2]`, `test_require_python_version[2]` reproduce on unmodified `main` and are unrelated to this change).
- Ruff clean.

**Other**

Mirrors the approach used in already-open security-fix PRs (#1983, #2026, #2035). Only new import is `shutil` (Python stdlib) — no `requirements.txt` change needed.

### DIFF
--- a/metagpt/utils/common.py
+++ b/metagpt/utils/common.py
@@ -22,8 +22,8 @@ import inspect
 import json
 import mimetypes
 import os
-import platform
 import re
+import shutil
 import sys
 import time
 import traceback
@@ -56,12 +56,9 @@ def check_cmd_exists(command) -> int:
     :param command: 待检查的命令
     :return: 如果命令存在，返回0，如果不存在，返回非0
     """
-    if platform.system().lower() == "windows":
-        check_command = "where " + command
-    else:
-        check_command = "command -v " + command + ' >/dev/null 2>&1 || { echo >&2 "no mermaid"; exit 1; }'
-    result = os.system(check_command)
-    return result
+    if not command:
+        return 1
+    return 0 if shutil.which(command) else 1
 
 
 def require_python_version(req_version: Tuple) -> bool:
--- a/metagpt/utils/mermaid.py
+++ b/metagpt/utils/mermaid.py
@@ -80,8 +80,8 @@ async def mermaid_to_file(
                 ]
             else:
                 commands = [config.mermaid.path, "-i", str(tmp), "-o", output_file, "-w", str(width), "-H", str(height)]
-            process = await asyncio.create_subprocess_shell(
-                " ".join(commands), stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
+            process = await asyncio.create_subprocess_exec(
+                *commands, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
             )
 
             stdout, stderr = await process.communicate()
--- a/tests/metagpt/utils/test_common.py
+++ b/tests/metagpt/utils/test_common.py
@@ -121,6 +121,25 @@ class TestGetProjectRoot:
             else:
                 assert result != 0
 
+    def test_check_cmd_exists_rejects_shell_injection(self, tmp_path):
+        """Regression test for issue #2037: command-injection via shell metacharacters.
+
+        check_cmd_exists must not invoke a shell on the input. Payloads containing
+        shell metacharacters must (a) return non-zero and (b) not execute the
+        smuggled command.
+        """
+        marker = tmp_path / "INJECTED"
+        payloads = [
+            "",
+            f"ls; touch {marker}",
+            f"ls && touch {marker}",
+            f"ls $(touch {marker})",
+            f"ls | tee {marker}",
+        ]
+        for payload in payloads:
+            assert check_cmd_exists(payload) != 0, f"payload should be rejected: {payload!r}"
+            assert not marker.exists(), f"shell injection executed for payload: {payload!r}"
+
     @pytest.mark.parametrize(("filename", "want"), [("1.md", "File list"), ("2.md", "Language"), ("3.md", "# TODOs")])
     @pytest.mark.asyncio
     async def test_parse_data_exception(self, filename, want):
--- a/tests/metagpt/utils/test_mermaid.py
+++ b/tests/metagpt/utils/test_mermaid.py
@@ -6,6 +6,10 @@
 @File    : test_mermaid.py
 """
 
+import os
+import sys
+from types import SimpleNamespace
+
 import pytest
 
 from metagpt.const import DEFAULT_WORKSPACE_ROOT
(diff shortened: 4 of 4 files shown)


GouravSingal-code:
@better629  can you please take a look at this pr

PR #2086, opened 2026-07-01
Add invinoveritas_review tool: independent review before irreversible actions

babyblueviper1:
## Summary

New tool, `invinoveritas_review` (`metagpt/tools/libs/invinoveritas_review.py`), registered in `tools/libs/__init__.py` alongside `terminal`/`editor`/`browser`/`deployer`/`git` — no other files changed.

## Why

`Terminal.forbidden_commands` is a small, hardcoded denylist (`"run dev"`, `"serve "` today). It's a good first line of defense, but it can only catch the exact patterns someone thought to list — anything else (a broad `rm -rf`, a destructive SQL migration, an unbounded `chmod`, a risky deploy config) passes through unchecked.

This tool is the general-purpose complement: sends the exact artifact (a shell command, a git diff, a config change) to [invinoveritas](https://api.babyblueviper.com) — an independent, model-agnostic reviewer — and returns a structured verdict (`approve` / `approve_with_concerns` / `reject`) + ranked issues, so an agent can check something novel before `Terminal.run_command` or `Deployer.deploy_to_public`, not just the two hardcoded strings.

Free registration (`/register`) returns an `api_key` with trial calls per tool; no crypto/payment setup needed to try it.

## Verified before submitting

Real calls against the live API, not mocks:

```
artifact: rm -rf /var/log/*.log && systemctl restart nginx
-> verdict: approve_with_concerns, confidence: 0.85
   "will delete all log files... could lead to loss of important log data"
```

```
INVINOVERITAS_API_KEY unset -> {"verdict": "unavailable", "confidence": 0.0,
"summary": "...register free...", "issues": []} — the documented fallback,
not an unhandled exception.
```

Ran the actual submitted function end-to-end (stubbed only `metagpt.tools.tool_registry.register_tool` so it imports without the full framework install — the `httpx` call and response-shaping logic is the real, unmodified code).

Happy to adjust naming/placement/style to match project conventions more closely.

### DIFF
--- a/metagpt/tools/libs/__init__.py
+++ b/metagpt/tools/libs/__init__.py
@@ -16,6 +16,7 @@ from metagpt.tools.libs import (
     browser,
     deployer,
     git,
+    invinoveritas_review,
 )
 from metagpt.tools.libs.env import get_env, set_get_env_entry, default_get_env, get_env_description, get_env_default
 
@@ -31,6 +32,7 @@ _ = (
     browser,
     deployer,
     git,
+    invinoveritas_review,
     get_env,
     get_env_default,
     get_env_description,
--- a/metagpt/tools/libs/invinoveritas_review.py
+++ b/metagpt/tools/libs/invinoveritas_review.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import annotations
+
+import os
+from typing import Optional
+
+import httpx
+
+from metagpt.tools.tool_registry import register_tool
+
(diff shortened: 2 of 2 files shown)


babyblueviper1:
Checking in after a month -- still mergeable on my end (single new file, no other changes touched), and the CI failure showing here is just the fork-PR deployment-protection gate expiring unapproved, not a real test failure (the separate build(3.9) run that didn't need approval passed). No pressure either way -- happy to update anything if the shape's not quite right for how tools/libs is organized these days.

PR #2142, opened 2026-08-27
fix(terminal): process newline-terminated markers

### LINKED ISSUE #2110: Terminal.run_command() hangs indefinitely on Linux/WSL2 — end-of-command marker never yielded from splitlines/tuple-unpack
### Summary

`metagpt.tools.libs.terminal.Terminal.run_command()` hangs indefinitely on Linux/WSL2 whenever the persistent shell's stdout ends with the end-of-command marker as the final line (i.e., always, under normal conditions).

The read loop in `_read_and_process_output` uses `*lines, tmp = output.splitlines(True)` after reading each byte, but when the buffer contains exactly one `\n`-terminated line, the tuple-unpacking idiom moves that last line into `tmp` instead of yielding it — so the marker check never fires and the loop awaits a next byte that will never come.

Because every `Engineer2._think()` call starts with `await self.terminal.run_command("pwd")`, this bug silently deadlocks the entire Engineer / DataAnalyst / RoleZero agent flow before any LLM call is made. The user sees "nothing happening" and eventually Ctrl-C's — no error message.

### Steps to reproduce

Any Linux or WSL2 host, current `main`:

```bash
pip install -e .
python - <<'PY'
import asyncio
from metagpt.tools.libs.terminal import Terminal

async def main():
    t = Terminal()
    out = await asyncio.wait_for(t.run_command("pwd"), timeout=10)
    print("got:", repr(out))

asyncio.run(main())
PY
```

Expected: prints the workspace path in under a second.
Actual: hangs 10s, `asyncio.TimeoutError`.

Running `metagpt "Create a hello world CLI"` exhibits the same hang right after the Team Leader delegates to Alex (Engineer) — logs stop at `Alex(Engineer) observed: [...]` and never advance.

### Root cause

`metagpt/tools/libs/terminal.py:_read_and_process_output`, current main:

```python
tmp = b""
while True:
    output = tmp + await self.process.stdout.read(1)
    if not output:
        continue
    *lines, tmp = output.splitlines(True)   # <-- BUG
    for line in lines:
        ...  # check for END_MARKER_VALUE
```

Minimal Python demonstration of the unpacking:

```python
>>> b"\x18\x19\x1b\x18\n".splitlines(True)
[b'\x18\x19\x1b\x18\n']
>>> *lines, tmp = [b'\x18\x19\x1b\x18\n']
>>> lines
[]
>>> tmp
b'\x18\x19\x1b\x18\n'
```

`splitlines(True)` on a single `\n`-terminated buffer returns a 1-element list. The unpacking places that element into `tmp` and leaves `lines` empty, so the marker line is never checked. The next `read(1)` blocks forever because bash has emitted everything and is waiting on stdin.

The exact byte sequence that triggers it in practice: workspace path + `\n` + `\x18\x19\x1b\x18\n` (the `END_MARKER_VALUE`). Byte-by-byte accumulation yields the workspace line normally (a following `\x18` pushes it out of `tmp`), but the final marker line is the last thing in the stream and stays stuck in `tmp` on the terminating `\n`.

### Suggested fix

Treat a trailing-newline buffer as complete instead of stashing it back into `tmp`:

```python
split = output.splitlines(True)
if split and split[-1].endswith((b"\n", b"\r")):
    lines = split
    tmp = b""
else:
    *lines, tmp = split
```

Locally patched with this and confirmed `metagpt "Create a Python CLI dice simulator"` runs end-to-end in ~2min instead of hanging forever.

Happy to open a PR if the maintainers agree with this approach.

### Environment

- MetaGPT: `main` @ 11cdf466d042aece04fc6cfd13b28e1a70341b1f
- Python: 3.10.12
- OS: Linux 6.6.87.2-microsoft-standard-WSL2 (Ubuntu on WSL2)
- Shell: `/usr/bin/bash` 5.x
- LLM backend: OpenAI-compatible local proxy (irrelevant — the hang is before any LLM call)

### Impact

Any Linux user of `Engineer2`, `DataAnalyst`, or any `RoleZero` subclass that touches `Terminal`. Silent hang, no traceback, no timeout — the process just idles until killed. The bug is particularly bad because there is no error to grep for; the only symptom is that the agent stops working after `_observe`.

ump45nose:
**Features**

- Process a complete stdout line immediately when its newline arrives instead of retaining it as an incomplete buffer.
- Add bounded waits and deterministic process cleanup to the existing terminal integration test.
- Fixes #2110. This is a smaller replacement for the self-closed #2111.

**Feature Docs**

Not applicable.

**Influence**

`Terminal.run_command()` no longer waits forever after bash emits the end-of-command marker as the final newline-terminated line.

**Result**

- Before: the real `Terminal.run_command("pwd")` path timed out after 3 seconds.
- After: the same bash-backed smoke test returned `"/tmp\n"` immediately.
- `ruff check metagpt/tools/libs/terminal.py tests/metagpt/tools/libs/test_terminal.py`
- `ruff format --check metagpt/tools/libs/terminal.py tests/metagpt/tools/libs/test_terminal.py`
- `git diff --check`

**Other**

The regression test uses `asyncio.wait_for` so a future marker-processing regression fails within five seconds instead of hanging the test job.

### DIFF
--- a/metagpt/tools/libs/terminal.py
+++ b/metagpt/tools/libs/terminal.py
@@ -157,7 +157,11 @@ class Terminal:
                 output = tmp + await self.process.stdout.read(1)
                 if not output:
                     continue
-                *lines, tmp = output.splitlines(True)
+                lines = output.splitlines(True)
+                if output.endswith(b"\n"):
+                    tmp = b""
+                else:
+                    *lines, tmp = lines
                 for line in lines:
                     line = line.decode(errors="ignore")
                     ix = line.rfind(END_MARKER_VALUE)
--- a/tests/metagpt/tools/libs/test_terminal.py
+++ b/tests/metagpt/tools/libs/test_terminal.py
@@ -1,3 +1,5 @@
+import asyncio
+
 import pytest
 
 from metagpt.const import DATA_PATH, METAGPT_ROOT
@@ -8,14 +10,17 @@ from metagpt.tools.libs.terminal import Terminal
 async def test_terminal():
     terminal = Terminal()
 
-    await terminal.run_command(f"cd {METAGPT_ROOT}")
-    output = await terminal.run_command("pwd")
-    assert output.strip() == str(METAGPT_ROOT)
+    try:
+        await asyncio.wait_for(terminal.run_command(f"cd {METAGPT_ROOT}"), timeout=5)
+        output = await asyncio.wait_for(terminal.run_command("pwd"), timeout=5)
+        assert output.strip() == str(METAGPT_ROOT)
 
-    # pwd now should be METAGPT_ROOT, cd data should land in DATA_PATH
-    await terminal.run_command("cd data")
-    output = await terminal.run_command("pwd")
-    assert output.strip() == str(DATA_PATH)
+        # pwd now should be METAGPT_ROOT, cd data should land in DATA_PATH
+        await asyncio.wait_for(terminal.run_command("cd data"), timeout=5)
+        output = await asyncio.wait_for(terminal.run_command("pwd"), timeout=5)
+        assert output.strip() == str(DATA_PATH)
+    finally:
+        await terminal.close()
 
 
 if __name__ == "__main__":


ump45nose:
The failing `Unit Tests / build (3.9)` job never reaches the test suite. `actions/checkout@v4` refuses to check out fork code from the repository\047s `pull_request_target` workflow, citing the trusted-token security boundary and requiring an explicit workflow-side decision (`allow-unsafe-pr-checkout: true` only after reviewing the risks). The companion Full Tests job is also still waiting. This cannot be fixed from the PR branch; could a maintainer update the workflow to test fork PRs safely, then rerun the checks?

